PR #11, opened 2016-03-04
Moving previous PR1726 to new repo (Account Disabling)

Pushpalanka:


Pushpalanka:
Keep onhold

Pushpalanka:
Ready for merge.. Have a know issue that "Enabling/Disabling users through admin service is not reflected on user profile UI view, but enabling/disabling functionality is happening fine". It should be ok for this PR to goto milestone pack with this, and fixing before alpha.

PR #22, opened 2016-03-15
Fix for 4395

ChamaraPhilipsuom:
Identity Server URL must be configured inside EndpointConfig.properties


components/authentication-framework/org.wso2.carbon.identity.application.authentication.endpoint.util/pom.xml
```diff
@@ -64,6 +64,11 @@
         </dependency>
         <dependency>
             <groupId>org.wso2.carbon.identity</groupId>
+            <artifactId>org.wso2.carbon.identity.application.authentication.framework</artifactId>
```
cdwijayarathna: It would be better if we can add getServicesUrl functionality to identity core util, so we can avoid adding this new dependency, and also get ServicesUrl would be possible to use in any identity component, WDYT?

components/authentication-framework/org.wso2.carbon.identity.application.authentication.endpoint.util/src/main/java/org/wso2/carbon/identity/application/authentication/endpoint/util/TenantDataManager.java
```diff
@@ -184,11 +179,11 @@ private static String buildFilePath(String path) throws IOException {
      * @return Property value
      */
     protected static String getPropertyValue(String key) {
-        if (key == Constants.SERVICES_URL && identityAvailable) {
-            return IdentityUtil.getServerURL(prop.getProperty(key),true, true);
-        } else {
-            return prop.getProperty(key);
+        if ((key == Constants.SERVICES_URL) && !prop.containsKey(Constants.SERVICES_URL)) {
```
thariyarox: you can use Constants.SERVICES_URL.equals(key) string comparison

PR #37, opened 2016-03-21
Improved the fix for IDENTITY-4395

ChamaraPhilipsuom:
Added improvement for the fix IDENTITY-4395


ChamaraPhilipsuom:
This is not needed in public.

PR #63, opened 2016-04-05
narrow down libthrift binding version range

harsha89:


harsha89:
In APIM we have installed MB features which packs libthrift 0.9.x versions. Due to existing version range identity components get binds to higher libthrift version which cause stub incompatibilities. Hence narrow down the version range.

pom.xml
```diff
@@ -1479,7 +1479,7 @@
         <neethi.osgi.version>2.0.4.wso2v4</neethi.osgi.version>
         <neethi.osgi.version.range>[2.0.4.wso2v4,3.0.0)</neethi.osgi.version.range>
         <libthrift.wso2.version>0.8.0.wso2v1</libthrift.wso2.version>
-        <libthrift.wso2.osgi.version.range>[0.8.0.wso2v1,1.0.0)</libthrift.wso2.osgi.version.range>
+        <libthrift.wso2.osgi.version.range>[0.8.0.wso2v1,0.9.0)</libthrift.wso2.osgi.version.range>
```
harsha89: In APIM we have installed MB features which packs libthrift 0.9.x versions. Due to existing version range identity components get binds to higher libthrift version which cause stub incompatibilities. Hence narrow down the version range.

PR #71, opened 2016-04-05
The operator ("<>") should be used.

AymanDF:
This pull request is focused on resolving occurrences of Sonar rule squid:S2293 - “The diamond operator ("<>") should be used ”. You can find more information about the issue here: 
https://dev.eclipse.org/sonar/rules/show/squid:S2293
Please let me know if you have any questions.
Ayman Abdelghany.


cdwijayarathna:
We can't use <> operator in jsp files.This cause syntax errors there. I think reason for this is we are still using java version lesser that 1.7 for jsp. Can you please fix those?

PR #80, opened 2016-04-07
Use tenant domain and user store domain in subject identifier default to false

madurangasiriwardena:


cdwijayarathna:
Can't we done this without introducing database migrations?

madurangasiriwardena:
Changed the fix without introducing database changes

PR #210, opened 2016-05-31
Improvements for UserStore config UI

kasungayan:


kasungayan:
This is already fixed with current CSRF improvements.

PR #220, opened 2016-06-05
Improvements required in framework

johannnallathamby:


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/inbound/IdentityMessageContext.java
```diff
@@ -37,6 +37,10 @@ public IdentityMessageContext(IdentityRequest request, Map<T1, T2> parameters) {
         this.request = request;
     }
 
+    public IdentityMessageContext(IdentityRequest request) {
+        this.request = request;
+    }
```
ChamaraPhilipsuom: We should call the super() constructor with parameters.

ChamaraPhilipsuom:
We need following line in FrameworkLoginResponseFactory.
`responseBuilder.setStatusCode(HttpServletResponse.SC_MOVED_TEMPORARILY);`
Otherwise sendRedirect()  will not work correctly in IdentityServlet.processHttpResponse.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/inbound/IdentityMessageContext.java
johannnallathamby: Didn't understand why we need to call super() with parameters. This constructor doesn't take Map like its overloaded version. So we can only call the no-arg constructor of its super class.

ChamaraPhilipsuom: yes we can call it. But there is a compile time error 

> error: constructor MessageContext in class MessageContext<T1,T2> cannot be applied to given types;

johannnallathamby: Fixed after offline discussion. Although still didn't understand how it can be overcome to call the no-arg constructor.

johannnallathamby: Reverted this fix. Found that it was due to not building identity.core

PR #238, opened 2016-06-09
Fixing 404 in policy editor UI - IDENTITY-4664

milindawickramasinghe:
Fix for https://wso2.org/jira/browse/IDENTITY-4664


omindu:
@milindawickramasinghe Thanks. Can you send another PR to release-productis-5.2.0 branch as well. The fix should be in both branches.

PR #294, opened 2016-06-13
improving idn_recovery_data table with adding primary and unique key constraints

IsuraD:


IsuraD:
close due to added "changing mapped Attributes of challengeQuestion1 and challegeQuestion2" with a seperate PR

PR #448, opened 2016-07-18
Improvements to IdentityValidateUtil class

mefarazath:
Adding check to validate whether provided whitelist/blacklist pattern keys are already registered and throw exceptions accordingly.

### DIFF
--- a/components/identity-core/org.wso2.carbon.identity.base/src/main/java/org/wso2/carbon/identity/base/IdentityValidationUtil.java
+++ b/components/identity-core/org.wso2.carbon.identity.base/src/main/java/org/wso2/carbon/identity/base/IdentityValidationUtil.java
@@ -32,6 +32,7 @@ public class IdentityValidationUtil {
     private static final String msgSection3 =
             "contains illegal characters matching one of the black list patterns [ %s ]";
     private static final String msgSection4 = " or ";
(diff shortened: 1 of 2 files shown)


components/identity-core/org.wso2.carbon.identity.base/src/main/java/org/wso2/carbon/identity/base/IdentityValidationUtil.java
```diff
@@ -76,8 +77,11 @@ public String getRegex() {
      * @param input             input
      * @param whiteListPatterns a String array of white list pattern keys
      * @return true if matches with any of the white list patterns
+     * @throws IdentityValidationException if a white list pattern key provided does not correspond to a registered
+     *                                     regex.
      */
-    public static boolean isValidOverWhiteListPatterns(String input, String... whiteListPatterns) {
+    public static boolean isValidOverWhiteListPatterns(String input, String... whiteListPatterns)
+            throws IdentityValidationException {
```
malithie: Well I think this exception should not be thrown here. Since it's the input being validated you can throw IllegalArgument exception instead. Throwing the checked exception will need to wrap up this method with a try catch clause.

mefarazath: fixed by throwing IllegalArgument exception.

PR #476, opened 2016-08-02
IDENTITY-4755 fixed.

thanujalk:


harsha1979:
This is already merged by GayanM as mentioned in above comment.

PR #526, opened 2016-08-21
IDENTITY-5008:New Feature:Assertion query profile-save assertions on database

gayangithub:
This pull request contain a new table structure for storing assertions on H2 database and the change of property file to store assertions.


pulasthi7:
Merged with #654

PR #537, opened 2016-08-24
Fix IDENTITY-4770

mefarazath:
* Removing ErrorInfo list from IdentityException class to get rid of WSDL errors encountered in workflow.


components/identity-core/org.wso2.carbon.identity.base/src/main/java/org/wso2/carbon/identity/base/IdentityException.java
```diff
@@ -33,109 +33,109 @@
 
     private static final long serialVersionUID = 725992116511551241L;
 
-    protected List<ErrorInfo> errorInfoList = new ArrayList<>();
+   // protected List<ErrorInfo> errorInfoList = new ArrayList<>();
```
mefarazath: fixed with https://github.com/wso2/carbon-identity-framework/pull/631

PR #558, opened 2016-08-30
Fixed IDENTITY-4832

IndunilRathnayake:


IndunilRathnayake:
Decided not to move the queries to identity-framework, but to remain in identity-inbound-auth-oauth
.

PR #614, opened 2016-10-06
Fixing the pom issues

pulasthi7:


pulasthi7:
Do not Merge, Requires further changes..

PR #698, opened 2016-11-24
IDENTITY-5277: Introduce Permissions for session management

ayshsandu:


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/resources/META-INF/services.xml
```diff
@@ -27,7 +27,7 @@
 		</parameter>
         <parameter name="adminService" locked="true">true</parameter>
         <parameter name="hiddenService" locked="true">true</parameter>
-        <parameter name="AuthorizationAction" locked="true">/permission/admin</parameter>
+        <parameter name="AuthorizationAction" locked="true">/permission/admin/manage/identity/authentication/session/delete</parameter>
```
darshanasbg: Even this service has only one operation better to follow our convention of defining permissions.

i.e.
Use Root permission of the resource as the permission of the service. In this case secure the service with,
`/permission/admin/manage/identity/authentication/session`

And use fine grained permission for each operations. In this case secure removeSession() method with,
`/permission/admin/manage/identity/authentication/session/delete`

In the end it would looks like follows,

```
        <parameter name="AuthorizationAction" locked="true">/permission/admin/manage/identity/authentication/session</parameter>

        <operation name="removeSession">
            <parameter name="AuthorizationAction" locked="true">/permission/admin/manage/identity/authentication/session/delete</parameter>
        </operation>
```

PR #850, opened 2017-01-23
IDENTITY-5575: Fixing the extention point tag name

nuwandiw:


mefarazath:
@pulasthi7 this needs to tracked in the documentation as well right?

pulasthi7:
@mefarazath Is this mentioned specifically in docs? @nuwandiw If so can you please do the needful?

PR #877, opened 2017-05-09
Encoding query params and disabling GET calls to update-policy-set.jsp

nuwandiw:


malithie:
What's the intention of restricting update-policy-set.jsp only for POSTs. As it seems this page is not changing the server state. It does not invoke a service operation but seems it's redirecting to some other pages for such operations. Shouldn't those be made ajax processors and restricted for POSTS. Also this respective page should initiate POST requests to them.

nuwandiw:
update-policy-set.jsp is restricted to be called on HTTP POST requests to make sure only trusted data is passed to other redirecting pages.

PR #878, opened 2017-05-15
Changing to use KeyProvider OSGi service 

ruwanta:


components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/KeyStoreManagerExtension.java
```diff
@@ -0,0 +1,50 @@
...
+     * Returns the Private Key for the given tenant domain.
+     *
+     * @param tenantDomain  the tenant domain
+     * @return
+     * @throws IdentityException upon any error accessing the underlying key store
+     */
+    Key getPrivateKey(String tenantDomain) throws IdentityException;
+
+    /**
+     * Returns the Public certificate for the given tenant domain.
```
harsha1979: and alias

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/KeyStoreManagerExtension.java
```diff
@@ -0,0 +1,50 @@
...
+     *
+     * @param tenantDomain  the tenant domain
+     * @param alias the alias used in the key-store to lookup the certificate.
+     * @return
```
harsha1979: add return type as well

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultKeyProviderService.java
```diff
@@ -0,0 +1,76 @@
...
+        this.defaultKeyStoreManagerExtension = defaultKeyStoreManagerExtension;
+    }
+
+    /**
+     * Updates the current used KeyStoreManagerExtension with the new extension.
+     * Updating with null value will cause the KeyProviderService to revert to using the default
+     * KeyStoreManagerExtension.
+     * @param keyStoreManagerExtension the new KeyStoreManagerExtension, or null to select the default one.
+     */
+    public void updateKeyStoreManagerExtension(KeyStoreManagerExtension keyStoreManagerExtension) {
```
harsha1979: Ideally, we don't need to keep to extension and we can replace the default one by new extension. Anyway, defaultKeyStoreManagerExtension is a private one and if some one update it, no one can use it again. So better to keep one holder and replace with update method.

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultKeyProviderService.java
```diff
@@ -0,0 +1,76 @@
...
+ */
+public class DefaultKeyProviderService implements KeyProviderService {
+
+    private static final Log log = LogFactory.getLog(DefaultKeyProviderService.class);
+
+    private KeyStoreManagerExtension keyStoreManagerExtension;
+    private KeyStoreManagerExtension defaultKeyStoreManagerExtension;
+
+    public DefaultKeyProviderService(KeyStoreManagerExtension defaultKeyStoreManagerExtension) {
+        this.defaultKeyStoreManagerExtension = defaultKeyStoreManagerExtension;
```
harsha1979: Better to check null here and throw exception by not allowing to create DefaultProviderService with null pass. Otherwise , other method in this class can be thrown null pointer exception

ruwanta: +1

PR #899, opened 2017-06-09
Fix IdentityConfigParser set wrong attribute of IdentityCookieConfig

yaosiang:
It seems cookieConfig set wrong attribute.


thanujalk:
@yaosiang Thanks for the PR.

PR #900, opened 2017-06-09
Change cookie attributes default value

yaosiang:
The default attribute value in CookieBuilder class is true. It means if user doesn't set particular value specifically, the attribute,  such as "Secure", "HttpOnly"  will be true.

Since identity server allow user to change these settings through identity.xml, the configuration file will be read and use in FrameworkUtils class.

Let's see following [code snippet](https://github.com/wso2/carbon-identity-framework/blob/master/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/util/FrameworkUtils.java#L554).

```
    if (commonAuthIdCookieConfig.isSecure()) {
        cookieBuilder.setSecure(commonAuthIdCookieConfig.isSecure());
    }
```
cookieBuilder.setSecure() will execute only when "Secure" attribute in identity.xml is true. It imply default value of "Secure" should be false.

### DIFF
--- a/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/CookieBuilder.java
+++ b/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/CookieBuilder.java
@@ -29,9 +29,9 @@ public class CookieBuilder {
     private String domain;
     private int maxAge = -1;
     private String path = "/";
-    private boolean secure = true;
+    private boolean secure = false;
     private int version = 0;
-    private boolean isHttpOnly = true;
+    private boolean isHttpOnly = false;
 
     public CookieBuilder(String name, String value) {
         this.name = name;


omindu:
@yaosiang Thank you for sending this. We'll merge this after reviewing.

thanujalk:
@yaosiang We need to mark these cookies "Secure" and "HttpOnly" by default considering the security concerns. 
If the system does not get compromised, any admin can uncomment following in the identity.xml and change the default behavior.
```
    <Cookies>
        <Cookie name="commonAuthId" domain="localhost" httpOnly="true" secure="true" />
    </Cookies>
```
Also following code only get executed when we have defined the above configuration.
```
    if (commonAuthIdCookieConfig.isSecure()) {
        cookieBuilder.setSecure(commonAuthIdCookieConfig.isSecure());
    }
```
So making these values false contradicts with our defult behaviour. So we need to keep them as it is.

yaosiang:
Yes, I agree marking these cookies "Secure" and "HttpOnly" by default considering the security concerns.

But, when any admin trying to uncomment following section in the identity.xml to change the default behavior. They will notice it doesn't works!

Suppose we uncomment these section in the identity.xml and change default behavior of secure (from true to false).
```
<Cookies>
    <Cookie name="commonAuthId" domain="localhost" httpOnly="true" secure="false" />
</Cookies>
```
The following if-condition will be evaluated when we have defined the above configuration. To make behavior actually changed, we also have to let ` cookieBuilder.setSecure` get executed. But, the secure value in the identity.xml is `false` now. It means `commonAuthIdCookieConfig.isSecure()` is false. Therefore, any code block in if-condition will not get executed. No behavior will be changed.
```
if (commonAuthIdCookieConfig.isSecure()) {
    cookieBuilder.setSecure(commonAuthIdCookieConfig.isSecure());
}
```
I suggest we can change the above if-condition in Framework Utils class instead of changing default value in the CookieBuilder class.

JKAUSHALYA:
Thanks for pointing out. We need to address this issue with a code level change. Hence created the following issue to track.

https://github.com/wso2/carbon-identity-framework/issues/1191

PR #924, opened 2017-06-30
Fix default role delete when app is deleted

vidurananayakkara:
Application/default role deletion when deleting store apps fixed


vidurananayakkara:
Closing pull request since this is not the proper fix

PR #926, opened 2017-07-04
Adding a util method to check user store supported operations.

hasinthaindrajee:


ashensw:
This is fixed in kernel 4.4.17. 
Refer: https://wso2.org/jira/browse/IDENTITY-6074

PR #936, opened 2017-07-17
IDENTITY-6153: Fix <OAuth2RevokeEPUrl> tag in identity.xml

piraveena:


piraveena:
@darshanasbg @mefarazath @madurangasiriwardena

rushmin:
Do we have any code block which reads the value of this configuration? If yes we need to change that code block as well.

piraveena:
Yes we are reading using [this](https://github.com/wso2/carbon-identity-framework/blob/master/components/identity-core/org.wso2.carbon.identity.base/src/main/java/org/wso2/carbon/identity/base/IdentityConstants.java#L311). Here the tag value is mentioned correctly.Only in the identity.xml, it was a mistake.

PR #977, opened 2017-08-02
IDENTITY-6200: Send custom error information from Authentication Framework

mefarazath:

With this improvement, we bring the capability to propagate any error information sent by an authenticator via the AuthenticationContext out of the Framework.

* Introduced populateErrorInformation() method as an extension point to customize the populating error information.
* Introduced constants for the Default Error detail related properties.

### Follow up actions

- Once this PR is merged and a component release is done, we need to refactor OAuth/SAML and other inbound authenticators to propogate the error details sent from Framework to their respective Relying Parties.

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultAuthenticationRequestHandler.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultAuthenticationRequestHandler.java
@@ -25,6 +25,7 @@ import org.apache.commons.logging.LogFactory;
 import org.wso2.carbon.CarbonConstants;
 import org.wso2.carbon.context.PrivilegedCarbonContext;
 import org.wso2.carbon.identity.application.authentication.framework.AuthenticationDataPublisher;
+import org.wso2.carbon.identity.application.authentication.framework.cache.AuthenticationResultCacheEntry;
 import org.wso2.carbon.identity.application.authentication.framework.config.model.ApplicationConfig;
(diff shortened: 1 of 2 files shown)


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultAuthenticationRequestHandler.java
```diff
@@ -508,4 +515,66 @@ protected boolean isPostAuthenticationExtensionCompleted(AuthenticationContext c
...
+    protected void populateErrorInformation(HttpServletRequest request,
+                                            HttpServletResponse response,
+                                            AuthenticationContext context) {
+
+        // get the authentication result
+        AuthenticationResult authenticationResult = getAuthenticationResult(request, response, context);
+
+        String errorCode = String.valueOf(context.getProperty(FrameworkConstants.AUTH_ERROR_CODE));
+        String errorMessage = String.valueOf(context.getProperty(FrameworkConstants.AUTH_ERROR_MSG));
+
```
danushkaf: Shouldn't we allow user to set username of the failure attempt. There could be use cases we will need that and it would be useful. I think better to allow that also. So someone who needs it can use it. In OAuth level we can set username to error_uri property.

mefarazath: Do we need to do this by default? 
In an authentication failure scenario, username could be invalid as well. Also, we are providing an extension point so I feel we should handle failed username in an extension by overriding this method.

PR #996, opened 2017-08-17
IDENTITY-6208: Possible Connection Leak in RegistryRecoveryDataStore

denuwanthi:

This is the fix for https://wso2.org/jira/browse/IDENTITY-6208.
Please refer the description in the jira.
In RegistryRecoveryDataStore.deleteOldResourcesIfFound() method, did not do any registry rollback transaction when an exception was thrown from registry.delete().
It catch the exception and logs the error. Therefore, the registry database connection is not closed.
It do the transaction roll back only if there is no resource in the registry path. It does not take into the account the time when registry operation fails.

-

ASAP

### Follow up actions

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/identity-mgt/org.wso2.carbon.identity.mgt/src/main/java/org/wso2/carbon/identity/mgt/store/RegistryRecoveryDataStore.java
+++ b/components/identity-mgt/org.wso2.carbon.identity.mgt/src/main/java/org/wso2/carbon/identity/mgt/store/RegistryRecoveryDataStore.java
@@ -276,25 +276,39 @@ public class RegistryRecoveryDataStore implements UserRecoveryDataStore {
             }
         }
 
+        boolean isTransactionSucceeded = false;
         try {
             if (collection != null) {
                 String[] resources = collection.getChildren();
                 for (String resource : resources) {
+                    isTransactionSucceeded = false;
                     String[] splittedResource = resource.split("___");
                     if (splittedResource.length == 3) {
                         //PRIMARY USER STORE
                         if (resource.contains("___" + userNameToValidate.toLowerCase() + "___")) {
-
-                            registry.beginTransaction();
-                            // Check whether the resource still exists for concurrent cases.
-                            if (registry.resourceExists(resource)) {
-                                registry.delete(resource);
-                                registry.commitTransaction();
-                            } else {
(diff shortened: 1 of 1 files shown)


components/identity-mgt/org.wso2.carbon.identity.mgt/src/main/java/org/wso2/carbon/identity/mgt/store/RegistryRecoveryDataStore.java
```diff
@@ -276,25 +276,39 @@ private void deleteOldResourcesIfFound(Registry registry, String userName, Strin
...
+                                    registry.delete(resource);
+                                    isTransactionSucceeded = true;
+                                } else {
+                                    // Already deleted by another thread. Do nothing.
+                                    if (log.isDebugEnabled()) {
+                                        log.debug("Confirmation code already deleted in path of resource : " + resource);
+                                    }
+                                }
+                            } catch (RegistryException ex) {
+                                log.error("Error while deleting the old confirmation code \n" + ex);
```
ashensw: This will print the ex object I guess. Shall we do this as `log.error("Error while deleting the old confirmation code", ex);`

denuwanthi: This will print the whole stracktrace.
Therefore will add a warn log to print the message and debug log to print the stacktrace.?

ashensw: Ok. That would be fine.

denuwanthi: The Throwable class has overridden the toString() method.
So, printing the exception does not print it as an object.
Therefore, I think we can keep the error log as it is.
Additionally will add a debug log to print the stacktrace.
Since this is in a for loop, printing the error stacktrace can consume the log file space.

PR #1156, opened 2017-11-07
Turn off log appending on unit tests, so that unit tests run bit faster.

ruwanta:

- Making the unit tests complete faster

ASAP

### Follow up actions

- None

#### General

#### Functionality

#### Code

### DIFF
--- a/test-utils/org.wso2.carbon.identity.testutil/src/main/java/org/wso2/carbon/identity/testutil/LogUtil.java
+++ b/test-utils/org.wso2.carbon.identity.testutil/src/main/java/org/wso2/carbon/identity/testutil/LogUtil.java
@@ -19,10 +19,9 @@
 
 package org.wso2.carbon.identity.testutil;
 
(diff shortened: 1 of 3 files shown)


test-utils/org.wso2.carbon.identity.testutil/src/main/java/org/wso2/carbon/identity/testutil/LogUtil.java
```diff
@@ -31,17 +30,13 @@
...
+
 	public static void configureAndAddConsoleAppender() {
-		//create an appender
-		ConsoleAppender console = new ConsoleAppender();
-		//configure the appender
-		String PATTERN = LOG_PATTERN;
-		console.setLayout(new PatternLayout(PATTERN));
-		console.activateOptions();
-		LogManager.getRootLogger().addAppender(console);
+        //create an appender
```
sagara-gunathunga: Since we have a comment here, shall we add the purpose of NullAppender ?

PR #1210, opened 2017-11-29
Fixed IDENTITY-6352

IndunilRathnayake:
Cannot edit existing workflow without changing workflow name


components/workflow-mgt/org.wso2.carbon.identity.workflow.mgt.ui/src/main/resources/web/workflow-mgt/add-wf-wizard.jsp
```diff
@@ -164,7 +164,9 @@
                 CARBON.showWarningDialog("<fmt:message key="workflow.error.empty.workflow.name"/>" , null ,null) ;
                 return false;
             }
-            if($.inArray(workflow_name, existingWorkflowNames) !== -1){
+
+            var workflowId = "\"<%=workflowId%>\"";
```
pulasthi7: Please check if the \" is necessary, it may result in a non empty id even if the actual Id is empty.

PR #1298, opened 2018-02-02
Fix for invalid authenticator name in multi-option steps

emswbandara:
Part of the fix for wso2/product-is#2245

When we are using the AuthenticationContext.getCurrentAuthenticator() method at the publisher proxy to get the authenticator name for multi option step configuration scenario this is evaluated as null since this value is explicitly set for only few scenarios.

With this fix, we pass the name as a property from abstract authenticator when calling the publisher proxy, and update the proxy to get the authenticator name from the property instead of context's current authenticator.


emswbandara:
@hasinthaindrajee @pulasthi7

PR #1299, opened 2018-02-02
Generalizing post authentication execution. 

hasinthaindrajee:

At the present we have post authentication criteria which are evaluated upon authentication in an authentication flow. Examples are "Handling missing mandatory claims" and "Authorization handling". According to the current implementation these logics are bind towards our framework implementation so that if we need to add a new post authentication evaluation criteria, we do not have an alternative other than changing framework source. 

With emerging requirements we may need to add more post authentication criteria in the future. For an example, we may need to intercept post authentication and request for consent on requested claims. Likewise there may be other requirements to intercept post authentication flow.

Foreseeing these requirements we are planing to generalize post authentication handling so that post authentication handling will no longer be a static part of framework. We should be able to add post authentication handlers as OSGI services. Upon this change, missing mandatory claim handler and authorization handler will be two OSGI post authentication handlers.    

[List all changes you want to add here. If you fixed an issue, please
add a reference to that issue as well.]

-

ASAP.

can merge this pull request.]

### Follow up actions
environments.]
Release a new version of xacml application authorization handler.
-


hasinthaindrajee:
Feature https://github.com/wso2/product-is/issues/2294

PR #1333, opened 2018-02-14
Add db scripts for xacml based scope validator for OAuth2 token

senthalan:


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/dbscripts/db2.sql
```diff
@@ -33,6 +33,13 @@ REFERENCING NEW AS NEW FOR EACH ROW MODE DB2SQL
     = (NEXTVAL FOR IDN_OAUTH_CONSUMER_APPS_SEQUENCE);
   END
 /
+CREATE TABLE IDN_OAUTH2_SCOPE_VALIDATORS (
+	APP_ID INTEGER NOT NULL,
+	SCOPE_VALIDATOR VARCHAR (128) NOT NULL,
```
Megala21: Are we sure SCOPE_VALIDATOR will be less than 128 characters? If not, is 128 is a safe limit?

senthalan: yes, as we are only saving the simple class name in scope_validator.

PR #1357, opened 2018-02-16
Modify identity scripts

denuwanthi:

-Modify all the identity scripts for each database type
-Increase the column size of CONSUMER_SECRET,ACCESS_TOKEN,REFRESH_TOKEN and AUTHORIZATION_CODE from 512 to 2048.
This is because with the introduction of new RSA+OAEP algorithm, the ciphertext adds the certificate to the ciphertext json representation. Since the he cert key size is now increased, the encrypted values of above oauth2 keys exceeds the 512 column size.  Thus, increased the column size.
-Add new columns ACCESS_TOKEN_HASH,REFRESH_TOKEN_HASH & AUTHORIZATION_CODE_HASH.
These columns will store the hash values of access token , refresh token and authorization code.
From now on we will always store a hash value of above keys when storing the origina value/encrypted value.
This will help when doing databse searches when dynamic encryption is enabled.
-Remove indexes for ACCESS_TOKEN & AUTHORIZATION_CODE.
This is because, we introduce a new hash column to do databse search queries.
Ex: A search query which used access token earlier is now searched using the hashed value of the access token. Therfore remove index for above columns
-Add new index for ACCESS_TOKEN_HASH and AUTHORIZATION_CODE_HASH 

This should be merged once following Kernel PRs are merged.
[1]https://github.com/wso2/carbon-kernel/pull/1698
[2]https://github.com/wso2/carbon-kernel/pull/1697

### Follow up actions

environments.]

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/dbscripts/mysql.sql
```diff
@@ -79,11 +81,11 @@ CREATE INDEX IDX_AT_CK_AU ON IDN_OAUTH2_ACCESS_TOKEN(CONSUMER_KEY_ID, AUTHZ_USER
 
 CREATE INDEX IDX_TC ON IDN_OAUTH2_ACCESS_TOKEN(TIME_CREATED);
 
-CREATE INDEX IDX_AT ON IDN_OAUTH2_ACCESS_TOKEN(ACCESS_TOKEN);
```
hasinthaindrajee: Why are we removing existing indexes ?

denuwanthi: With the introduction of hash column, the search operations will be done using the hash value.
Therefore removed the old  index

PR #1361, opened 2018-02-19
support tenancy for self signup

hasinthaindrajee:

[List all changes you want to add here. If you fixed an issue, please
add a reference to that issue as well.]

- Support tenancy for self sign up


hasinthaindrajee:
Fixed with https://github.com/wso2-extensions/identity-carbon-auth-rest/pull/56

PR #1420, opened 2018-03-04
Fix wso2/product-is#2724: Introduced a global config to disable consent management for SSO

omindu:

The following config is introduced from this PR in identity.xml to enable/disable consent management during SSO. The feature will be enabled by default.

```xml
<Consent>
       <!--Specify whether consent management should be enable during SSO.-->
       <EnableSSOConsentManagement>true</EnableSSOConsentManagement>
</Consent>
```

When invoking the `SSOConsentManagementService` if the above property is set to `false`, an `SSOConsentDisabledException` will be thrown.

This PR also introduces the `SSOConsentDisabledException` and `SSOConsentServiceException` exception types for service specific error handling.

Immediately

### Follow up actions

@mefarazath Update the OAuth endpoint according to the new service changes.


sherenewso2:
Updated doc [1] 

[1] https://docs.wso2.com/display/IS550/Consent+Management+with+Single-Sign-On

PR #1528, opened 2018-04-19
Fix no error messages display when reusing the redirection URL and when using the redirection URL after the expiry time

DMHP:

Fix no error messages display when reusing the redirection URL and when using the redirection URL after the expiry time

After merging https://github.com/wso2-extensions/identity-governance/pull/223/files

### DIFF
--- a/components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/java/org/wso2/carbon/identity/mgt/endpoint/client/api/NotificationApi.java
+++ b/components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/java/org/wso2/carbon/identity/mgt/endpoint/client/api/NotificationApi.java
@@ -228,6 +228,10 @@ public class NotificationApi {
      */
     public void validateCodePostCall(CodeValidationRequest code) throws ApiException {
 
+        // verify the required parameter 'code' is set
+        if (code == null) {
+            throw new ApiException(400, "Missing the required parameter 'code' when calling validateCodePost(Async)");
+        }
         List<Pair> queryParams = new ArrayList<>();
         Map<String, String> headerParams = new HashMap<>();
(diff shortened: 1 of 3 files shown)


components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/java/org/wso2/carbon/identity/mgt/endpoint/client/api/NotificationApi.java
```diff
@@ -229,7 +217,64 @@ public void setPasswordPost(ResetPasswordRequest resetPasswordRequest) throws Ap
...
+        };
+        final String headerAccept = apiClient.selectHeaderAccept(contentType);
+        final String headerContentType = apiClient.selectHeaderContentType(contentType);
+        String tenantDomain = getTenantDomain(code);
+        setBasePath(tenantDomain);
+
+        if (headerAccept != null) {
+            headerParams.put("Accept", headerAccept);
+        }
+        headerParams.put("Content-Type", headerContentType);
```
malithie: What is the reason to specifically check for null only for headerAccept.
Looks like both had been initiated over contentType variable defined above.

malithie: Also, define the variable along with its usage so it improves readability as it is within the context being used

DMHP: headerAccept returns null if the contentType.length is 0. So we need to check null for headerAccept.  headerContentType returns a string if the contentType.length is 0. SO no need to check null here for headerContentType.

PR #1535, opened 2018-04-24
Provide support to get user's role list in conditional auth js

pulasthi7:
Fixes https://github.com/wso2/product-is/issues/3110


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/model/graph/js/JsAuthenticatedUser.java
```diff
@@ -120,4 +132,20 @@ public boolean hasMember(String name) {
...
+
+    private String[] getLocalRoles() {
+
+        if (idp == null || FrameworkConstants.LOCAL.equals(idp)) {
+            RealmService realmService = FrameworkServiceDataHolder.getInstance().getRealmService();
+            int usersTenantId = IdentityTenantUtil.getTenantId(getWrapped().getTenantDomain());
+
+            try {
+                UserRealm userRealm = realmService.getTenantUserRealm(usersTenantId);
+                return userRealm.getUserStoreManager().getRoleListOfUser(getWrapped().getUserName());
```
mefarazath: Shouldn't we have the userstore domain appended to the username?

ruwanta: @mefarazath , then it would be ned method. We do not want to append to the name.
e.f.
var d = user.domain;
var r = user.roles;

mefarazath: I didn't understand "ned"? 
BTW my concern was that how will this method work for a user belonging to a secondary user store.

PR #1631, opened 2018-06-12
Add prompt for wait while calling outside endpoints in authentication

madurangasiriwardena:

From the adaptive authentication, we can call outside analytics engine to decide next authentication steps. Getting the response from outside analytics engine may take some time due to the delays in the analytics engine. During this period,
- user will get a frozen page
- current thread which is executing the authentication is blocked

To address the situation, we planned below flow.
- When the authentication framework initiate the call to the external analytics engine, we redirect the user to a waiting page.
- From the waiting page we periodically get the status of the external call we have initiated. For this we are using a servlet and ajax to get the status.
- When the external call is completed, redirect the user to /commonauth again.
- Waiting page has a timeout and it will redirect the user to a error page after the timeout.

This behavior is disabled by default. We can enable this behavior from the identity.xml with the property "PromptOnLongWait"


components/authentication-framework/org.wso2.carbon.identity.application.authentication.endpoint.util/src/main/java/org/wso2/carbon/identity/application/authentication/endpoint/util/AdaptiveAuthUtil.java
```diff
@@ -0,0 +1,126 @@
...
+            if (configFile.exists()) {
+                adaptiveAuthConfigs.put("AdaptiveAuth.HTTPConnectionTimeout", IdentityUtil.getProperty("AdaptiveAuth.HTTPConnectionTimeout"));
+                adaptiveAuthConfigs.put("AdaptiveAuth.RefreshInterval", IdentityUtil.getProperty("AdaptiveAuth.RefreshInterval"));
+            } else {
+                Properties props = new Properties();
+                try (InputStream inputStream = AdaptiveAuthUtil.class.getClassLoader()
+                        .getResourceAsStream(Constants.TenantConstants.CONFIG_FILE_NAME)) {
+                    if (inputStream != null) {
+                        props.load(inputStream);
+                        log.debug(Constants.TenantConstants.CONFIG_FILE_NAME
```
pulasthi7: better wrap inside a debug enabled check

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/dbscripts/db2.sql
```diff
@@ -541,6 +541,27 @@ CREATE TRIGGER SP_AUTH_SCRIPT_TRIG NO CASCADE
                     SET (NEW.ID) = (NEXTVAL FOR SP_AUTH_SCRIPT_SEQ);
                 END
 /
+CREATE TABLE SP_WAIT_STATUS (
```
pulasthi7: Shall we have this with prefix IDN_AUTH

PR #1635, opened 2018-06-13
Product-IS Issue #2192 Fix

danAtInv:
- Fix for #https://github.com/wso2/product-is/issues/2192
- FrameworkUtils.java - add a method to find previously authenticated RequestPathAuthenticators
- DefaultStepHandler.java - In DefaultStepHandler::handle(), call the previous method if no authenticated StepAuthenticators were found

- no preconditions

### Follow up actions
- no follow-up actions

#### General

- [x] **Is this PR explained thoroughly?** All code changes must be accounted for in the PR description.
- [x] **Is the PR labeled correctly?**

#### Functionality

- [x] **Are all requirements met?** Compare implemented functionality with the requirements specification.
- [x] **Does the UI work as expected?** There should be no Javascript errors in the console; all resources should load. There should be no unexpected errors. Deliberately try to break the feature to find out if there are corner cases that are not handled.

#### Code

- [x] **Do you fully understand the introduced changes to the code?** If not ask for clarification, it might uncover ways to solve a problem in a more elegant and efficient way.
- [x] **Does the PR introduce any inefficient database requests?** Use the debug server to check for duplicate requests.
- [x] **Are all necessary strings marked for translation?** All strings that are exposed to users via the UI must be [marked for translation](https://docs.djangoproject.com/en/1.10/topics/i18n/translation/).

#### Tests

- [x] **Are there sufficient test cases?** Ensure that all components are tested individually; models, forms, and serializers should be tested in isolation even if a test for a view covers these components.
- I will work on a test case - in the meantime if you can review the code that would be great.
- [x] **If this is a new feature or a significant change to an existing feature?** has the manual testing spreadsheet been updated with instructions for manual testing?

#### Security

- [x] **Confirm this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets.**
- [x] **Are all UI and API inputs run through forms or serializers?**
- [x] **Are all external inputs validated and sanitized appropriately?**
- [x] **Does all branching logic have a default case?**
- [x] **Does this solution handle outliers and edge cases gracefully?**
- [x] **Are all external communications secured and restricted to SSL?**

#### Documentation

- [x ] **Are changes to the UI documented in the platform docs?** If this PR introduces new platform site functionality or changes existing ones, the changes should be documented.
- [x] **Are changes to the API documented in the API docs?** If this PR introduces new API functionality or changes existing ones, the changes must be documented.
- [x] **Are reusable components documented?** If this PR introduces components that are relevant to other developers (for instance a mixin for a view or a generic form) they should be documented in the Wiki.

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/util/FrameworkUtils.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/util/FrameworkUtils.java
@@ -1011,6 +1011,72 @@ public class FrameworkUtils {
         return idpAuthenticatorMap;
     }
 
+    // IDS_2192 - Add a method to retrieve RequestPath Authenticators who have previously authenticated the user
+    public static Map<String, AuthenticatorConfig> getAuthenticatedRequestPathIdPs(SequenceConfig sequenceConfig, Map<String, AuthenticatedIdPData> authenticatedIdPs) {
+        if (log.isDebugEnabled()) {
+            log.debug("Finding already authenticated request path IdPs");
+        }
+
+        Map<String, AuthenticatorConfig> idpAuthenticatorMap = new HashMap<String, AuthenticatorConfig>();
+        List<AuthenticatorConfig> requestPathAuthenticators = sequenceConfig.getReqPathAuthenticators();
+
+        if (authenticatedIdPs != null && !authenticatedIdPs.isEmpty()) {
+
+            for (AuthenticatorConfig authenticatorConfig : requestPathAuthenticators) {
+
+                if (log.isDebugEnabled()) {
+                    log.debug(String.format("Considering the authenticator '%s'", authenticatorConfig.getName()));
+                }
+
+                String authenticatorName = authenticatorConfig.getName();
+
+                if (log.isDebugEnabled()) {
+                    log.debug(String.format("%d IdP(s) found in the step.", authenticatedIdPs.size()));
+                }
+
+                for (Map.Entry<String, AuthenticatedIdPData> entry : authenticatedIdPs.entrySet()) {
+                    String idpName = entry.getKey();
+                    AuthenticatedIdPData authenticatedIdPData = entry.getValue();
+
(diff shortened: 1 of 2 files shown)


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/util/FrameworkUtils.java
```diff
@@ -1011,6 +1011,72 @@ public static String getQueryStringWithFrameworkContextId(String originalQuerySt
...
     }
 
+    // IDS_2192 - Add a method to retrieve RequestPath Authenticators who have previously authenticated the user
+    public static Map<String, AuthenticatorConfig> getAuthenticatedRequestPathIdPs(SequenceConfig sequenceConfig, Map<String, AuthenticatedIdPData> authenticatedIdPs) {
+        if (log.isDebugEnabled()) {
+            log.debug("Finding already authenticated request path IdPs");
+        }
+
+        Map<String, AuthenticatorConfig> idpAuthenticatorMap = new HashMap<String, AuthenticatorConfig>();
+        List<AuthenticatorConfig> requestPathAuthenticators = sequenceConfig.getReqPathAuthenticators();
```
malithie: This will work only if the same request path authenticator has been configured for each service provider that will be participating in the user session.
So that, when the first application uses request path authentication, the second application which will use the basic authentication scheme, but also has request path authentication option being configured will work.
If the request path authenticator has not been configured for the second application, it should still prompt for credentials in this fix.

We need to clearly validate the user story for this usecase.  That is whether the user should get SSO for an application that will only use basic authentication scheme, when the user is already logged into an application over request path basic authentication scheme.

As I see, with this fix, additional configurations needs to be done to achieve above.

malithie:
Hi @danAtInv,

Thanks a lot for contributing.
Due to the above mentioned concern we will be closing this PR.
We have initiated a discussion at dev@wso2.org [1] to decide on the appropriate fix for this issue.
You can follow up on the solution via the thread.

[1] https://mail.wso2.org/mailarchive/dev/2018-July/077707.html

danAtInv:
Thanks @malithie - I will follow that thread!

PR #1729, opened 2018-07-20
Add a new property to the identity.xml to enable all the details.

GDRDABARERA:

Adding a new property to the SCIM tag of identity.xml

Resolves wso2/product-is#3447


nilasini:
Please don't forget to create a doc issue to mention about this property in scim document

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml
```diff
@@ -519,6 +519,7 @@
         -->
         <UserEPUrl>${carbon.protocol}://${carbon.host}:${carbon.management.port}/wso2/scim/Users</UserEPUrl>
         <GroupEPUrl>${carbon.protocol}://${carbon.host}:${carbon.management.port}/wso2/scim/Groups</GroupEPUrl>
+        <ShowAllUserDetails>false</ShowAllUserDetails>
```
nilasini: Seems like we have to change the identity.xml tag name from "ShowAllUserDetails" to something else because name it self-mentioning that we are going to show all the user details but actually we are going to show only one user details. Shall we use a more appropriate name?

PR #1794, opened 2018-08-09
Set maxlength value of description field for Service Provider Configuration section

SwatiNEC:
Set maxlength value of description field(Add service provider and edit service provider configuration page).
In current scenario, If we enter more than 1024 characters in description field then it will generate exception in logs while inserting data in database. 
Refer the below issue 
https://github.com/wso2/product-is/issues/2040


madurangasiriwardena:
@SwatiNEC Thanks you for your contribution!

PR #1899, opened 2018-09-18
Fixing 'Something went wrong during the authentication process.Please try signing in again' for invalid credentials

nuwandiw:
Fixes https://github.com/wso2/product-is/issues/3711

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/AbstractApplicationAuthenticator.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/AbstractApplicationAuthenticator.java
@@ -96,9 +96,9 @@ public abstract class AbstractApplicationAuthenticator implements ApplicationAut
                     // Decide whether we need to redirect to the login page to retry authentication.
                     boolean sendToMultiOptionPage =
                             isStepHasMultiOption(context) && isRedirectToMultiOptionPageOnFailure();
+                    context.setRetrying(retryAuthenticationEnabled());
                     if (retryAuthenticationEnabled(context) && !sendToMultiOptionPage) {
                         // The Authenticator will re-initiate the authentication and retry.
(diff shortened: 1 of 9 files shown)


ruwanta:
@nuwandiw can you please explain the approach how this solves the issue?

nuwandiw:
> @nuwandiw can you please explain the approach how this solves the issue?

@ruwanta mentioned issue occurs because when using GraphBasedSequenceHandler, retry attempt is not prompted (Retrying is skipped from the framework level because retry flow would start without going through the script, therefore mentioned conditions would not be applicable to the retry flow). 

When user provides invalid credentials, as retrying is skipped, DefaultStepHandler returns to the GraphBasedSequenceHandler with the FlowStatus FAIL_COMPLETED. As we do not have **onFail** functions implemented in js templates, authentication is failed without any specific error parameters. Hence giving "Something went wrong during the authentication process.Please try signing in again" error.

In the provided solution.
1. In the case of authentication failure from the authenticator level (not the script), I have added the **isRetrying** variable value to the context (which will indicated whether retrying is enabled in the particular failed authenticator). At that point, even though retrying is skipped from framework, context has the information whether particular authenticator allows retrying.
2. In the provided templates, I have added **onFail** function where it checks whether retrying is allowed and call the same onLoginRequest(context) accordingly.

So at the case of invalid credentials, it will execute **onFail** function and perform retrying accordingly.

features/application-mgt/org.wso2.carbon.identity.application.mgt.server.feature/resources/identity/authenticationTemplates/age.json
```diff
@@ -68,6 +68,10 @@
     "                Log.debug('User ' + context.currentKnownSubject.identifier + ' is under aged. Hence denied to login.');",
     "                sendError(errorPage, errorPageParameters);",
     "            }",
+    "        }, onFail: function (context) {",
```
ruwanta: Can we skip this and handle it at the 
```
case FAIL_COMPLETED:
```
when there is no such function registered. 
i.e. the default behavior is to handle retry if retry is enabled.

So the script is simple for this case.

PR #1984, opened 2018-11-16
Change encode type of user input values for IDP

isharailanga:

This PR is to fix XSS Issues found in IS for IDP management by encoding userIdClaimURI and roleClaimURI.

### DIFF
--- a/components/idp-mgt/org.wso2.carbon.idp.mgt.ui/src/main/resources/web/idpmgt/idp-mgt-edit.jsp
+++ b/components/idp-mgt/org.wso2.carbon.idp.mgt.ui/src/main/resources/web/idpmgt/idp-mgt-edit.jsp
@@ -290,8 +290,8 @@
 
         identityProviderClaims = identityProvider.getClaimConfig().getIdpClaims();
 
-        userIdClaimURI = Encode.forHtmlContent(identityProvider.getClaimConfig().getUserClaimURI());
-        roleClaimURI = Encode.forHtmlContent(identityProvider.getClaimConfig().getRoleClaimURI());
+        userIdClaimURI = identityProvider.getClaimConfig().getUserClaimURI();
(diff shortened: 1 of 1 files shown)


components/idp-mgt/org.wso2.carbon.idp.mgt.ui/src/main/resources/web/idpmgt/idp-mgt-edit.jsp
```diff
@@ -290,8 +290,8 @@
 
         identityProviderClaims = identityProvider.getClaimConfig().getIdpClaims();
 
-        userIdClaimURI = identityProvider.getClaimConfig().getUserClaimURI();
-        roleClaimURI = identityProvider.getClaimConfig().getRoleClaimURI();
+        userIdClaimURI = Encode.forHtmlContent(identityProvider.getClaimConfig().getUserClaimURI());
```
thariyarox: Do not encode and store the encoded value in the variable. Do the encoding at the time the value is displayed in the page. Because, encoding should be different based on the place where it is displayed (like element value, element attribute, hyperlink URL etc.)

thariyarox: For example, userIdClaimURI is added to the page source as below.

if ('<%=userIdClaimURI%>' == '') {

Above is inside a javascript block. So the correct fix is below.

if ('<%=Encode.forJavaScript(userIdClaimURI)%>' == '') {

isharailanga: Fixed in 1e2222e9efdc0ad2025ad5dec052fcc71bfb3bd4

PR #2051, opened 2019-01-25
 [configuration-store] Refactor date-time Metadata

Kavindu-Dodan:
- Introduce 'created' meta-data
- Store DateTime metadata as UTC timestamps in the db
- Reformat both date time metadata parameters to ISO 8601 compliant UTC standard
- Backward compatibility for scenarios without 'created' metadata in the database
- Fix date-time metadata parameters in the created/replaced API operation responses


tharindu-bandara:
Fixes https://github.com/wso2/product-is/issues/4346

PR #2075, opened 2019-02-21
Validate callback URL in recovery requests

Abilashini:
This PR contains the changes to validate call back URL in the username recovery request and manage call back URL validation when callback URL is null during password reset.

### DIFF
--- a/components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/java/org/wso2/carbon/identity/mgt/endpoint/client/CallBackValidator.java
+++ b/components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/java/org/wso2/carbon/identity/mgt/endpoint/client/CallBackValidator.java
@@ -83,8 +83,8 @@ public class CallBackValidator {
         String callbackRegex = null;
         if (idpProperties != null) {
             for (IdentityProviderProperty property : idpProperties) {
-                if (property.getName().equals(IdentityManagementEndpointConstants.UserInfoRecovery
-                        .RECOVERY_CALLBACK_REGEX)) {
+                if (IdentityManagementEndpointConstants.UserInfoRecovery.RECOVERY_CALLBACK_REGEX.equals(property
+                        .getValue())) {
                     callbackRegex = property.getValue();
                     if (log.isDebugEnabled()) {
                         log.debug("Configured recovery callback URL regex: " + callbackRegex);
--- a/components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/java/org/wso2/carbon/identity/mgt/endpoint/client/CallBackValidator.java
+++ b/components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/java/org/wso2/carbon/identity/mgt/endpoint/client/CallBackValidator.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.carbon.identity.mgt.endpoint.client;
+
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.logging.Log;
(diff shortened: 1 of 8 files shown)


components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/java/org/wso2/carbon/identity/mgt/endpoint/client/CallBackValidator.java
```diff
@@ -0,0 +1,112 @@
...
+     * @param tenantDomain of the user
+     * @return the status of the validation
+     * @throws IdentityRecoveryException
+     */
+    public boolean isValidCallbackURL(String callbackURL, String tenantDomain, boolean isUserPortalURL)
+            throws IdentityRecoveryException {
+
+        if (isUserPortalURL) {
+            if (log.isDebugEnabled()) {
+                log.debug("Callback URL is equal to the user portal URL: " + isUserPortalURL);
```
emswbandara: this debug will be logged only if isUserPortalURL is true. Shall we add a debug log to trace the false scenario?

components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/java/org/wso2/carbon/identity/mgt/endpoint/client/CallBackValidator.java
```diff
@@ -0,0 +1,112 @@
...
+        } else {
+            if (log.isDebugEnabled()) {
+                log.debug("Resident identity provider is not found for the tenant domain: " + tenantDomain);
+            }
+        }
+
+        String callbackRegex = null;
+        if (idpProperties != null) {
+            for (IdentityProviderProperty property : idpProperties) {
+                if (property.getName().equals(IdentityManagementEndpointConstants.UserInfoRecovery
```
emswbandara: shall we switch the two arguments of equals statement since RECOVERY_CALLBACK_REGEX is a constant here? This will avoid possible NPEs.

components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/webapp/password-reset-complete.jsp
```diff
@@ -51,7 +58,12 @@
         property.setKey("callback");
         property.setValue(URLEncoder.encode(callback, "UTF-8"));
         properties.add(property);
-        
+
+        Property userPortalURLProperty = new Property();
+        userPortalURLProperty.setKey("isUserPortalURL");
+        userPortalURLProperty.setValue(String.valueOf(isUserPortalURL));
```
emswbandara: shall we use boolean.parseboolean(isUserPortalURL) here?

components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/webapp/username-recovery.jsp
```diff
@@ -208,8 +208,7 @@
                             <%}%>
 
                             <%
-                                String callback = Encode.forHtmlAttribute
-                                        (request.getParameter("callback"));
+                                String callback = request.getParameter("callback");
```
emswbandara: Why was  Encode.forHtmlAttribute removed?

Abilashini: when callback is null, Encode.forHtmlAttribute will return the value as "null", which was affecting the following line where user portal URL will be assigned to the call back when it is null.

components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/webapp/password-reset-complete.jsp
Abilashini: isUserPortalURL is already a boolean value. Why do you expect it to be parsed into a boolean value again? Hence userPortalURLProperty.setValue() method accepts only String value, I have wrapped the isUserPortalURL variable into a String.

components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/java/org/wso2/carbon/identity/mgt/endpoint/client/CallBackValidator.java
Abilashini: user portal URL will be assigned to the callback URL by the recovery endpoint only when callback URL is null. isURLPortalURL will be passed to the server only to notify that the callback URL is equal to userPortalURL to avoid callback validation since this will be handled internally by the server. Else case will be logged at https://github.com/wso2/carbon-identity-framework/pull/2075/files/799f4304c4a890d0e0d0e880e9325f54640ecf62#diff-a0ef0b46165988efc2915a3ae61f2b70R103.

components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint/src/main/java/org/wso2/carbon/identity/mgt/endpoint/client/CallBackValidator.java
Abilashini: fixed

PR #2088, opened 2019-03-07
Fix invalid username passed to view-users.jsp from view-roles.jsp page

mefarazath:
$subject


tharinduwijewardane:
Fixes: product-ei#3316

PR #2132, opened 2019-03-27
Fix issue of initializeRegistry() in IdentityTenantUtil is called for every SSO request

GDRDABARERA:
Resolves wso2/product-is#3076

### DIFF
--- a/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityTenantUtil.java
+++ b/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityTenantUtil.java
@@ -45,6 +45,7 @@ import org.wso2.carbon.utils.multitenancy.MultitenantUtils;
 
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpSession;
+import java.util.concurrent.ConcurrentHashMap;
 
 public class IdentityTenantUtil {
 
@@ -53,6 +54,7 @@ public class IdentityTenantUtil {
     private static Log log = LogFactory.getLog(IdentityTenantUtil.class);
     private static TenantRegistryLoader tenantRegistryLoader;
     private static BundleContext bundleContext;
+    protected static ConcurrentHashMap<Integer,Boolean> tenantIdMap = new ConcurrentHashMap<Integer,Boolean>();
 
     public static TenantRegistryLoader getTenantRegistryLoader() {
         return tenantRegistryLoader;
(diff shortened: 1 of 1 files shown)


components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityTenantUtil.java
```diff
@@ -199,36 +201,40 @@ public static void setBundleContext(BundleContext bundleContext) {
...
-                carbonContext.setTenantDomain(tenantDomain, true);
-                BundleContext bundleContext = IdentityTenantUtil.getBundleContext();
-                if (bundleContext != null) {
-                    ServiceTracker tracker = new ServiceTracker(bundleContext, AuthenticationObserver.class.getName(), null);
-                    tracker.open();
-                    Object[] services = tracker.getServices();
-                    if (services != null) {
-                        for (Object service : services) {
-                            ((AuthenticationObserver) service).startedAuthentication(tenantId);
+        if (tenantIdMap.get(tenantId) == null || Boolean.valueOf(!tenantIdMap.get(tenantId))) {
```
malithie: What's the reason for this boolean condition
Tenant id is an int value right. Won't this always be false. Why convert for a boolean

malithie: ! should come before Boolean.valueOf()

malithie: Still when invocations happens concurrently there's a chance to get the tenants loaded. Isn't it

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityTenantUtil.java
```diff
@@ -53,6 +54,7 @@
     private static Log log = LogFactory.getLog(IdentityTenantUtil.class);
     private static TenantRegistryLoader tenantRegistryLoader;
     private static BundleContext bundleContext;
+    protected static ConcurrentHashMap<Integer,Boolean> tenantIdMap = new ConcurrentHashMap<Integer,Boolean>();
```
Kavindu-Dodan: Improve formatting

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityTenantUtil.java
Kavindu-Dodan: Can if conditions in line 204 and 206 combined ?

malithie: Just noted that the hash map includes a boolean as a value. Why a conversion is needed here

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityTenantUtil.java
malithie: What's the intention of using a protected variable here

malithie:
What would happen when tenant unloads. 
AFAIK, that would unload tenant registry yet, with this fix as it has been tracked that the tenant has loaded it won't load the tenant again right

PR #2152, opened 2019-04-01
Fix #4772:Role based provisioning not works as expected

IsurangaPerera:
- Fix wso2/product-is#4772

**Scenario: Role-based provisioning is enabled and only users with role_x are permitted to be provisioned**

When a user is created with role_x assigned then the preAddUser event is triggered and the provisioning process initiates. At this stage, existing implementation tries to fetch all user roles from the user store manager. As user roles are not yet persisted only **internal/everyone** role is applied to the user being provisioned. Hence role-based provisioning not works for the user even though he has intended roles assigned to him.

In this fix we consider roles from the provisioning entity as well as roles fetched from the user store.

can merge this pull request.]

### Follow up actions

environments.]

-

#### General

- [ ] **Is the PR labelled correctly?**

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/provisioning/org.wso2.carbon.identity.provisioning/src/main/java/org/wso2/carbon/identity/provisioning/OutboundProvisioningManager.java
+++ b/components/provisioning/org.wso2.carbon.identity.provisioning/src/main/java/org/wso2/carbon/identity/provisioning/OutboundProvisioningManager.java
@@ -19,6 +19,7 @@
 package org.wso2.carbon.identity.provisioning;
 
 import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.collections.ListUtils;
 import org.apache.commons.collections.MapUtils;
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.logging.Log;
@@ -68,6 +69,8 @@ import java.util.Map.Entry;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 /**
  *
@@ -824,11 +827,18 @@ public class OutboundProvisioningManager {
             return true;
         }
 
-        String userName = getUserName(provisioningEntity.getAttributes());
-        List<String> roleListOfUser = getUserRoles(userName, tenantDomain);
+        if (provisioningEntity.getAttributes() != null) {
+            String userName = getUserName(provisioningEntity.getAttributes());
+            List<String> provisioningRoleList = Arrays.asList(provisionByRoleList);
 
-        for (String provisionByRole : provisionByRoleList) {
-            if (roleListOfUser.contains(provisionByRole)) {
+            List<String> roleListOfUser = getUserRoles(userName, tenantDomain);
+            if (userHasProvisioningRoles(roleListOfUser, provisioningRoleList, userName)) {
+                return true;
+            }
+            List<String> newRoleListOfUser = provisioningEntity.getAttributes().get(ClaimMapping.build
+                        (IdentityProvisioningConstants.GROUP_CLAIM_URI, null, null, false));
+
+            if (userHasProvisioningRoles(newRoleListOfUser, provisioningRoleList, userName)) {
                 return true;
             }
         }
@@ -1051,5 +1061,28 @@ public class OutboundProvisioningManager {
         if (mappedGroups != null && !mappedGroups.isEmpty()) {
             ProvisioningUtil.setClaimValue(groupAttributeName, provisioningEntity.getAttributes(), mappedGroups);
         }
+    }
+
+    /**
+     * Check if user roles has common elements with the provisioning role list
+     *
+     * @param userRoles
+     * @param provisioningRoles
+     * @param userName
+     */
+    private boolean userHasProvisioningRoles(List<String> userRoles, List<String> provisioningRoles, String userName) {
+
+        if (CollectionUtils.isNotEmpty(userRoles) && CollectionUtils.isNotEmpty(provisioningRoles)) {
+            for(String provisioningRole : provisioningRoles) {
+                if (userRoles.contains(provisioningRole)) {
+                    if (log.isDebugEnabled()) {
+                        log.debug("User with userName : " + userName + " has provisioning role(s) assigned.");
+                    }
+                    return true;
+                }
+            }
+        }
+
+        return false;
     }
 }


emswbandara:
@IsurangaPerera could you please update the description on the problem observed and the approach taken to resolve the case

components/provisioning/org.wso2.carbon.identity.provisioning/src/main/java/org/wso2/carbon/identity/provisioning/OutboundProvisioningManager.java
```diff
@@ -825,7 +827,10 @@ protected boolean canUserBeProvisioned(ProvisioningEntity provisioningEntity,
         }
 
         String userName = getUserName(provisioningEntity.getAttributes());
-        List<String> roleListOfUser = getUserRoles(userName, tenantDomain);
+        List<String> roleListOfUser = Stream.concat(getUserRoles(userName, tenantDomain).stream(),
+                provisioningEntity.getAttributes().get(ClaimMapping.build
```
ruwanta: what if 
provisioningEntity.getAttributes() is null ?
provisioningEntity.getAttributes().get(x) is null ?

components/provisioning/org.wso2.carbon.identity.provisioning/src/main/java/org/wso2/carbon/identity/provisioning/OutboundProvisioningManager.java
```diff
@@ -824,12 +826,22 @@ protected boolean canUserBeProvisioned(ProvisioningEntity provisioningEntity,
...
+            String userName = getUserName(provisioningEntity.getAttributes());
+            List<String> roleListOfUser = getUserRoles(userName, tenantDomain);
+
+            List<String> newRoleListOfUser = provisioningEntity.getAttributes().get(ClaimMapping.build
+                    (IdentityProvisioningConstants.GROUP_CLAIM_URI, null, null, false));
 
-        for (String provisionByRole : provisionByRoleList) {
-            if (roleListOfUser.contains(provisionByRole)) {
-                return true;
+            if(newRoleListOfUser != null) {
```
malithie: Apply formatting

components/provisioning/org.wso2.carbon.identity.provisioning/src/main/java/org/wso2/carbon/identity/provisioning/OutboundProvisioningManager.java
```diff
@@ -824,12 +826,22 @@ protected boolean canUserBeProvisioned(ProvisioningEntity provisioningEntity,
             return true;
         }
 
-        String userName = getUserName(provisioningEntity.getAttributes());
-        List<String> roleListOfUser = getUserRoles(userName, tenantDomain);
+        if(provisioningEntity.getAttributes() != null) {
```
emswbandara: formatting issue

components/provisioning/org.wso2.carbon.identity.provisioning/src/main/java/org/wso2/carbon/identity/provisioning/OutboundProvisioningManager.java
```diff
@@ -824,12 +826,22 @@ protected boolean canUserBeProvisioned(ProvisioningEntity provisioningEntity,
...
-        for (String provisionByRole : provisionByRoleList) {
-            if (roleListOfUser.contains(provisionByRole)) {
-                return true;
+            if(newRoleListOfUser != null) {
+                 roleListOfUser = Stream.concat(roleListOfUser.stream(), newRoleListOfUser.stream()).distinct()
+                         .collect(Collectors.toList());
+            }
+
+            for (String provisionByRole : provisionByRoleList) {
+                if (roleListOfUser.contains(provisionByRole)) {
```
emswbandara: shall we use ListUtils here?

components/provisioning/org.wso2.carbon.identity.provisioning/src/main/java/org/wso2/carbon/identity/provisioning/OutboundProvisioningManager.java
```diff
@@ -824,12 +826,22 @@ protected boolean canUserBeProvisioned(ProvisioningEntity provisioningEntity,
...
+                    return true;
```
emswbandara: shall we add a debug log here?

components/provisioning/org.wso2.carbon.identity.provisioning/src/main/java/org/wso2/carbon/identity/provisioning/OutboundProvisioningManager.java
ruwanta: We are just checking a particular role is available in two another list?
If so, it is not good to create another list with distinct strings. This creates large list, and add garbage pressure.
Best is to check on original two lists one after another, which does not create new list.

components/provisioning/org.wso2.carbon.identity.provisioning/src/main/java/org/wso2/carbon/identity/provisioning/OutboundProvisioningManager.java
```diff
@@ -1052,4 +1059,23 @@ private void updateMappedGroupForAttribute(ProvisioningEntity provisioningEntity
...
+     *
+     * @param userRoles
+     * @param provisioningRoles
+     */
+    private boolean userHasProvisioningRoles(List<String> userRoles, List<String> provisioningRoles, String userName) {
+
+        if (CollectionUtils.isNotEmpty(userRoles) && CollectionUtils.isNotEmpty(provisioningRoles) && !ListUtils
+                .intersection(userRoles, Arrays.asList(provisioningRoles)).isEmpty()) {
+            if (log.isDebugEnabled()) {
+                log.debug("User with userName " + userName + " has provisioning role(s) assigned.");
```
ruwanta: General pattern for log is like this(please note colon)
  log.debug("User with userName : " + userName + " has provisioning role(s) assigned.");

Reason, lets assume a user having a name "not found.". Can you imagine what is the log line printed.

ruwanta: Also 
ListUtils.intersection() still creates another which is unnecessary for the purpose

PR #2179, opened 2019-04-10
Improved token and code cleanup logic.

dmxunlimit:
Improved token and code cleanup logic and fix the possible continues looping issue on the chunk table.

N/A
-

N/A

### Follow up actions

N/A

-

#### General

N/A

#### Functionality

N/A
#### Code

N/A
#### Tests

N/A
#### Security

N/A
#### Documentation

N/A


madurangasiriwardena:
@dmxunlimit can you please explain the improvement you have done?

dmxunlimit:
> @dmxunlimit can you please explain the improvement you have done?

Initially, the **Expired** tokens were deleted based on the token and refresh token validity period calculation. However, later noticed that if the token state is Expired, there is no need to verify the validity period again, Hence it's improved the logic to delete the Expired tokens without checking for the validity period.

PR #2186, opened 2019-04-11
Fix #5009:JIT provisioning password policy validation is enabled, backend throws a error for a incorrect password But UI proceed the user for the next step

IsurangaPerera:
- Fix wso2/product-is#5009
This will redirect the user to an error page if JIT provisioning failed due to attribute(ex. password) validation error. 

- ASAP

### Follow up actions
- N/A


ashensw:
This is merged to resolve the critical state of the issue. This will be immediately addressed with 5.8.0-beta2.

PR #2192, opened 2019-04-19
Remove policy publisher executor

madurangasiriwardena:

- Remove policy publisher executor and publish policies in the same thread.


darshanasbg:
whats the reason for the change?

madurangasiriwardena:
Policies are published from a pool of 2 threads. When there are multiple publish requests comes, this thread pool gets busy and policy publishing gets delayed. But the client can't get the status of the policy publishing and the initial requests for the XACML policy may provide incorrect responses since the policy is not deployed yet.
I can't see any reason for the policy to be published in a separate thread.

PR #2196, opened 2019-04-22
AuthenticatedUser constructor fails to add federated IDP name

emswbandara:
Part of fix for https://github.com/wso2/product-is/issues/4879
Depends on the OAuth fix : https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/1097


madurangasiriwardena:
@emswbandara can you add the link to the OAuth fix?

PR #2207, opened 2019-04-29
Add an error log message

DMHP:
Add an error log message as the exception is wrapped from the above level and the message is modified.


components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/ApplicationMgtValidator.java
```diff
@@ -85,6 +85,7 @@ public void validateSPConfigurations(ServiceProvider serviceProvider, String ten
         validateRoleConfigs(validationMsg, serviceProvider.getPermissionAndRoleConfig(), tenantDomain);
 
         if (!validationMsg.isEmpty()) {
+            log.error("SP Configurations validation failed: " + Arrays.toString(validationMsg.toArray(new String[0])));
```
malithie: Can we add more context information on the SP.
Moreover, it seems an error is thrown, isn't this handled in the service layer

PR #2245, opened 2019-06-13
Protect Scope endpoint

mefarazath:
Currently, the API[1] to add scope bindings is unprotected. This fix adds the necessary configurations to protect it.

[1] https://docs.wso2.com/display/IS570/Using+the+OAuth2+Scopes+REST+APIs


madurangasiriwardena:
We need to identify the correct permission level for this endpoint.

mefarazath:
> We need to identify the correct permission level for this endpoint.

Created https://github.com/wso2/product-is/issues/5762 to track this

PR #2267, opened 2019-06-26
Add httpStatusCode value to IdentityException

ayshsandu:

Introduce HTTP status code as a property in IdentityException.


ruwanta:
this is Identity core. It needs not have any notion of HTTP.

PR #2290, opened 2019-07-15
deduplicate header in authenticationendpoint

hauntingEcho:
deduplicate the header bar in authenticationendpoint. This simplifies skinning & corrections.

### Follow up actions
deduplication across all webapps would be ideal, but I'm not yet familiar enough with the overall codebase to do so.


jeradrutnam:
Thanks for the PR. Same change has been done with a separate effort and moved the webapp to a new location (https://github.com/wso2/identity-apps/tree/master/apps/authentication-portal/src/main/webapp).

PR #2294, opened 2019-07-17
Issue 5696 fixes

VivekVinushanth:
Issue 5696 : https://github.com/wso2/product-is/issues/5696

Method signature changed as required.
tenant domain can be taken from tenantId,

### DIFF
--- a/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityTenantUtil.java
+++ b/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityTenantUtil.java
@@ -196,10 +196,22 @@ public class IdentityTenantUtil {
         IdentityTenantUtil.bundleContext = bundleContext;
     }
 
+    /**
+     * @deprecated
+     * initializeRegistry(int tenantId, String tenantDomain)
+     *
+     * method signature changed.Will be removed fro next version.
+     * Because, tenant domain can be taken from tenantId
+     * Use ,
+     * initializeRegistry(int tenantId)
+     *
+     */
+
+    @Deprecated
+    public static void initializeRegistry(int tenantId, String tenantDomain) {}
 
     public static void initializeRegistry(int tenantId) throws IdentityException {
         String tenantDomain=getTenantDomain(tenantId);
-
 
         if (tenantId != MultitenantConstants.SUPER_TENANT_ID) {
             try {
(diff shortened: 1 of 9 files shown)


components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityTenantUtil.java
```diff
@@ -188,16 +194,21 @@ public static void setRealmService(RealmService realmService) {
...
+     *
+     */
+    @Deprecated
+    public static void initializeRegistry(int tenantId, String tenantDomain) throws IdentityException {
+        initializeRegistry(tenantId);
     }
 
-
-    public static void initializeRegistry(int tenantId, String tenantDomain) throws IdentityException {
+    public static void initializeRegistry(int tenantId) throws IdentityException {
```
janakamarasena: Let's not modify the method signature of an existing public method. Instead, add an overloaded method and call it from within the old method

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityTenantUtil.java
```diff
@@ -58,11 +58,14 @@ public static TenantRegistryLoader getTenantRegistryLoader() {
         return tenantRegistryLoader;
     }
 
+    public static void setTenantRegistryLoader(TenantRegistryLoader tenantRegistryLoader) {
```
ruwanta: From where this method is called?
It is not safe to set a static variable via public method as anyone can set to any value , thus system can break or can lead to race conditions.

madurangasiriwardena: This seems to be an existing method. After the refactoring, it is showing as a new method. At the moment it is called from the service component when it gets a refference to `TenantRegistryLoader` service instance.

ruwanta: ack

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityTenantUtil.java
```diff
@@ -164,13 +167,16 @@ public static RegistryService getRegistryService() {
         return registryService;
     }
 
+    public static void setRegistryService(RegistryService registryService) {
```
GANGANI, line 170: Add a new line after public methods. Check all the places. Let's add method comments for public methods.

VivekVinushanth: Fixed with https://github.com/wso2/carbon-identity-framework/pull/3767

PR #2300, opened 2019-07-17
remove unused assets from accountrecoveryendpoint

hauntingEcho:
everything uses the copies in the libs folder instead

removes unused copies of libraries, to make working with the code easier.  Moves the two remaining non-library js files from the assets folder to the js folder.


jeradrutnam:
Thank you very much for your contribution. We have manually added your changes to the webapp, as the webapp is being moved to a new repo [1] and no longer maintain here. 

[1] https://github.com/wso2/identity-apps/commit/d64252e2c2f76af23bbbb5fa887cb4310e89c64e

PR #2317, opened 2019-07-23
Fix issues in the IDP search

senthalan:

Fixes https://github.com/wso2/product-is/issues/5965 and fix issues when searching
Along with this fix, it makes the IDP search similar to the SP search


components/idp-mgt/org.wso2.carbon.idp.mgt/src/main/java/org/wso2/carbon/idp/mgt/dao/IdPManagementDAO.java
```diff
@@ -176,9 +176,17 @@
 	try {
 		String sqlStmt = IdPManagementConstants.SQLQueries.GET_IDPS_NAME_SQL;
 		prepStmt = dbConnection.prepareStatement(sqlStmt);
+        if (StringUtils.isNotBlank(filter)) {
```
IsuraD: Looks like this is not formatted

PR #2332, opened 2019-07-29
Fixing issue 5941: Adding support for federated users attributes in XACML

piraveena:
Fixes wso2/product-is#5941

As an improvement for the above issue, a flag will be set in the PEPs for federated users. If that flag is set, we check in DefaultAttributeFinder, whether the flag is coming in XACML request from PEP and prevent from pulling attributes from userstore if the flag is set.

In order to have backword compatibility, some people may already use federated users with JIT provisioning and have the XACML policy to read the userattributes from userstore. In order have the backwardcompatibility, intorduced a config in EntitlementProperties config file.

If the below config is true,
`org.wso2.carbon.identity.entitlement.pip.DefaultAttributeFinder.1=MapFederatedUsersToLocal,true` 
Then for federated users also , attribute values can be taken from local userstore. Else, PIP will not pull the attributes from userstore. If it is false, it will skip the local userstore. **By default this is enabled**

In order to skip userstore, According to the new config model, in the deployment.toml file it should be as follows:
```
[identity.entitlement.default_attribute_finder.properties]
MapFederatedUsersToLocal=false
```

can merge this pull request.]

### Follow up actions

environments.]

-

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/entitlement/org.wso2.carbon.identity.entitlement/src/main/java/org/wso2/carbon/identity/entitlement/pip/DefaultAttributeFinder.java
+++ b/components/entitlement/org.wso2.carbon.identity.entitlement/src/main/java/org/wso2/carbon/identity/entitlement/pip/DefaultAttributeFinder.java
@@ -50,7 +50,7 @@ public class DefaultAttributeFinder extends AbstractPIPAttributeFinder {
 
     private static Log log = LogFactory.getLog(DefaultAttributeFinder.class);
     private Set<String> supportedAttrs = new HashSet<String>();
-    private boolean mapFederatedUsersToLocal = true;
+    private boolean mapFederatedUsersToLocal = false;
     private static final String MAP_FEDERATED_USERS_TO_LOCAL = "MapFederatedUsersToLocal";
 
     /**
--- a/components/entitlement/org.wso2.carbon.identity.entitlement/src/main/java/org/wso2/carbon/identity/entitlement/pip/DefaultAttributeFinder.java
+++ b/components/entitlement/org.wso2.carbon.identity.entitlement/src/main/java/org/wso2/carbon/identity/entitlement/pip/DefaultAttributeFinder.java
@@ -21,14 +21,21 @@ package org.wso2.carbon.identity.entitlement.pip;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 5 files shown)


components/entitlement/org.wso2.carbon.identity.entitlement/src/main/java/org/wso2/carbon/identity/entitlement/pip/DefaultAttributeFinder.java
```diff
@@ -60,7 +71,46 @@ public String getModuleName() {
...
+        Set<String> values = null;
+        EvaluationResult userType = evaluationCtx.getAttribute(new URI(StringAttribute.identifier), new URI(
+                PDPConstants.USER_TYPE_ID), issuer, new URI(PDPConstants.USER_CATEGORY));
+        String userTypeId = null;
+        if (userType != null && userType.getAttributeValue() != null && userType.getAttributeValue().isBag()) {
+            BagAttribute bagAttribute = (BagAttribute) userType.getAttributeValue();
+            if (bagAttribute.size() > 0) {
+                userTypeId = ((AttributeValue) bagAttribute.iterator().next()).encode();
+                if (log.isDebugEnabled()) {
+                    log.debug(String.format("The user is a %1$s user", userTypeId));
```
omindu: Is it possible to include the username in the log ?

ruwanta: It is better to have following to avoid confusion in the log
String.format("The user type of the user is %s", userTypeId)
otherwise the log is printed like
```
The user is a locals user
```

components/entitlement/org.wso2.carbon.identity.entitlement/src/main/java/org/wso2/carbon/identity/entitlement/pip/DefaultAttributeFinder.java
```diff
@@ -60,7 +71,46 @@ public String getModuleName() {
         return "Default Attribute Finder";
     }
 
-    /*
+    /**
+     * This method is introduced in order to check whether the user is a local or federated and if it is a
+     * federated user, prevent from obtaining user attributes from userstore.
```
nilasini: fix grammatical error

PR #2356, opened 2019-08-07
Fix local claims getting listed as null in adaptive auth script when there is an existing session

emswbandara:
Fixes wso2/product-is#6079

with [1] we have added a fix to include Identity Provider name for the currentKnownSubject for all scenarios. With this fix when there is an already authenticated session by the 1st SP , and when the second SP tries to login, the user's local claims becomes null in the adaptive authentication script. This is because, in [2] when creating a JSClaims Object, as IDP name is now available it will always go to the method [2] instead of [3]. In that method the JSAuthenticatedUser is retrieved by getting the currentAuthenticatedIPS from authenticationContext. This is null for the second SP. As a possible solution, if the currentAuthenticatedIDPs is null, we can retrieve the StepConfig of subject identifier Step and retrieve the authenticated user.

[1] https://github.com/wso2/carbon-identity-framework/pull/2188
[2]

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/model/graph/js/JsClaims.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/model/graph/js/JsClaims.java
@@ -26,6 +26,7 @@ import org.wso2.carbon.context.PrivilegedCarbonContext;
 import org.wso2.carbon.identity.application.authentication.framework.ApplicationAuthenticator;
 import org.wso2.carbon.identity.application.authentication.framework.config.ConfigurationFacade;
 import org.wso2.carbon.identity.application.authentication.framework.config.model.ExternalIdPConfig;
+import org.wso2.carbon.identity.application.authentication.framework.config.model.StepConfig;
 import org.wso2.carbon.identity.application.authentication.framework.context.AuthenticationContext;
 import org.wso2.carbon.identity.application.authentication.framework.internal.FrameworkServiceDataHolder;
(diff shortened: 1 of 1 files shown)


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/model/graph/js/JsClaims.java
```diff
@@ -87,6 +89,46 @@ public void initializeContext(AuthenticationContext context) {
...
+        }
+        Map<Integer, StepConfig> stepConfigs = getContext().getSequenceConfig().getAuthenticationGraph().getStepMap();
+        Optional<StepConfig> subjectIdentifierStep = stepConfigs.values().stream()
+                .filter(stepConfig -> (stepConfig.isCompleted() && stepConfig.isSubjectIdentifierStep())).findFirst();
+        if (subjectIdentifierStep.isPresent()) {
+            return subjectIdentifierStep.get();
+        } else if (getContext().getCurrentStep() > 0) {
+            return stepConfigs.get(getContext().getCurrentStep());
+        } else {
+            return null;
```
JKAUSHALYA: This will set the authenticated user as a null. Is that fine? For example, then the [1] and [2] will return null and false.
[1] public Object getMember(String claimUri)
[2] public boolean hasMember(String claimUri)

emswbandara: this method will be called, if there is an already existing session for an AuthenticatedUser. which means that user will already have a subject identifier step executed in previous session. There AFAIU if the subject identifier step is null , it would be okay to return null object.. WDYT?

PR #2411, opened 2019-09-09
Add active session limit handler param configs to default.json and key-mappings.json

gayashanbc:
Related https://github.com/wso2-extensions/identity-local-auth-basicauth/pull/61


senthalan:
Please add this configuration to the application-authenticators.xml file as well as we support the old config model as well.

gayashanbc:
> Please add this configuration to the application-authenticators.xml file as well as we support the old config model as well.

@senthalan Added the Session Handler config params to `application-authentication.xml` to support the old config model.

PR #2450, opened 2019-09-26
Update js method signature

senthalan:

Update js method singature

Fix https://github.com/wso2/product-is/issues/6303, https://github.com/wso2/product-is/issues/6296

After this merge the doc change PR : https://github.com/wso2/docs-is/pull/418


ruwanta:
@senthalan Would all javascript needed to change to 
```
var onLoginRequest() 
```
for all authentication request ?

senthalan:
@ruwanta, The issue was not with the `onLoginRequest`. The 
`function onLoginRequest(context) {` worked in java 11. But the issue was with the child functions which are called inside onLoginRequest. 

 But we thought of changing that as well to make consistency in the script

PR #2462, opened 2019-09-28
Use a mask value for encrypted properties instead of a caching approach

tharindu-bandara:
**This pull request fixes the https://github.com/wso2/product-is/issues/6410.**

### Use Case:
1. Secondary user store configurations can be modified with the UserStoreConfigAdminService. A client is calling this admin service to modify secondary user store configurations. 
2. The client retrieves secondary user store configurations from the get method in the admin service.
3. Then the client modifies the received payload, sends the edit request to the IS.

### Existing Implementation:
- The secondary user store configurations contain the password of the user store, along with any other property which marked as encrypted to true.
- The encrypted fields are masked(with a random phrase) in the payload by IS, in step 02/03 above. The masking is done as follows.
-- Each encrypted field is stored in a RandomPassword object. This stores the random password phrase and the password value against the property name.
-- The set of RandomPassword objects for a user store is then stored in the RandomPasswordContainer object with a UUID. Then this RandomPasswordContainer is stored in the RandomPasswordContainerCache against the same UUID value.
-- The UUID value is sent in the get response in step 02, as a user store property. All the encrypted properties in the response have the random phrase as it’s value.
-- When the edit request received during step 03, IS will repopulate all the encrypted values from the RandomPasswordContainerCache instead of the random phrase, using the same UUID value received. 
-- If any of the random phrases are changed by the client, then this is interpreted as a change of the property. Then the new value is encrypted and stored.

### An Issue With A Clustered Deployment:
- The client request user store configurations(during step 02) and the request is processed by the node 01(This will also generate the above-discussed UUID).
- The client sends the edit request(during step 03) but now the request is processed by the node 02.
- The RandomPasswordContainerCache is not handled with Hazelcast clustering thus not shared through the cluster(as we do not enable distributed caching). Then node 02 does not retrieve a RandomPasswordContainer for the received UUID eventually failing the user store update request.

### Solution Introduced With This Pull Request:
- We will be removing the RandomPasswordContainer and the UUID based approach for masking encrypted fields. Instead, a constant value is used as a masking value for an encrypted field.
- During the update request(step 03), the encrypted fields are retrieved from the user realm. 
- If any encrypted field in the update request is equal to the same constant value we used as the mask, then that field is not modified by the client. In such cases, retrieved encrypted values from the user realm are used to replace the masking constant value in the corresponding field.

The UUID value is redundant with this approach. But it will be kept in the payload in order to preserve backward compatibility.

### DIFF
--- a/components/user-store/org.wso2.carbon.identity.user.store.configuration/src/main/java/org/wso2/carbon/identity/user/store/configuration/utils/SecondaryUserStoreConfigurationUtil.java
+++ b/components/user-store/org.wso2.carbon.identity.user.store.configuration/src/main/java/org/wso2/carbon/identity/user/store/configuration/utils/SecondaryUserStoreConfigurationUtil.java
@@ -27,11 +27,10 @@ import org.w3c.dom.Element;
 import org.w3c.dom.NodeList;
 import org.wso2.carbon.base.api.ServerConfigurationService;
 import org.wso2.carbon.context.CarbonContext;
+import org.wso2.carbon.context.PrivilegedCarbonContext;
 import org.wso2.carbon.core.util.CryptoUtil;
 import org.wso2.carbon.identity.core.util.IdentityUtil;
-import org.wso2.carbon.identity.user.store.configuration.beans.RandomPassword;
-import org.wso2.carbon.identity.user.store.configuration.beans.RandomPasswordContainer;
-import org.wso2.carbon.identity.user.store.configuration.cache.RandomPasswordContainerCache;
+import org.wso2.carbon.identity.user.store.configuration.beans.MaskedProperty;
 import org.wso2.carbon.identity.user.store.configuration.dao.UserStoreDAO;
 import org.wso2.carbon.identity.user.store.configuration.dao.impl.FileBasedUserStoreDAOFactory;
 import org.wso2.carbon.identity.user.store.configuration.dto.PropertyDTO;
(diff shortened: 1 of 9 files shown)


components/user-store/org.wso2.carbon.identity.user.store.configuration/src/main/java/org/wso2/carbon/identity/user/store/configuration/utils/SecondaryUserStoreConfigurationUtil.java
```diff
@@ -320,6 +322,24 @@ public static String getUserStoreProperties(UserStoreDTO userStoreDTO) throws Id
...
+        Map<String, String> secondaryUserStoreProperties = null;
+        try {
+            UserStoreManager secondaryUserStoreManager = getSecondaryUserStoreManager(userStoreDomain);
+            if (secondaryUserStoreManager != null) {
+                secondaryUserStoreProperties = secondaryUserStoreManager.getRealmConfiguration()
+                        .getUserStoreProperties();
+            }
+        } catch (UserStoreException e) {
+            String errorMessage = "Error while retrieving user store configurations";
+            log.error(errorMessage, e);
```
DMHP: If the throwing message is not overriding from a higher level , we don't need to log and throw.

tharindu-bandara: Fixed in https://github.com/wso2/carbon-identity-framework/commit/97cd4b67b5b86fdf62f343d31c78b02f74d52063

PR #2484, opened 2019-10-22
Service Provider API backend changes

mefarazath:


mefarazath:
Closed since this approach was not used.

PR #2507, opened 2019-10-30
Adding improvements to application management using resoureId

mefarazath:
* Prevent renaming an app to system default local sp's name
* Prevent deleting the local sp
* Add constants


mefarazath:
part of https://github.com/wso2/product-is/issues/6766

PR #2508, opened 2019-10-31
Added two jsp pages to device flow support

ayeshajay:
**Purpose**
Since device flow needs a page to insert user code and instruct the user at the end of the flow, two new jap pages needed.


ayeshajay:
Send a PR to identity apps https://github.com/wso2/identity-apps/pull/172

PR #2519, opened 2019-11-05
API changes to application resource management APIs

mefarazath:
* Include user performing action as a parameter for state-changing operation APIs
* Remove user performing action parameter for GET operations in the ApplicationResourceManagementListener
* Change create application API to return the created application information

### DIFF
--- a/components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/ApplicationManagementService.java
+++ b/components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/ApplicationManagementService.java
@@ -46,24 +46,6 @@ public abstract class ApplicationManagementService implements ApplicationPaginat
     }
 
     /**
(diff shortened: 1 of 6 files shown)


components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/ApplicationManagementService.java
```diff
@@ -45,24 +45,6 @@ public static ApplicationManagementService getInstance() {
...
-     * user will assigned to the created role.Internal roles used.
-     *
-     * @param serviceProvider Service Provider Name
-     * @param tenantDomain Tenant Domain
-     * @param username User Name
-     * @return
-     * @throws IdentityApplicationManagementException
-     * @deprecated  This method is replaced by {@link #createApplicationWithTemplate}
-     */
-    @Deprecated
```
mefarazath: We are removing this API. 

But any client calling this API won't break since we have introduced a new API with the same signature that returns the created application.

PR #2522, opened 2019-11-06
Add new methods to ApplicationManagementService

vihanga-liyanage:


components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/ApplicationManagementService.java
```diff
@@ -346,5 +354,12 @@ public abstract boolean isExistingApplicationTemplate(String templateName, Strin
         return new ApplicationBasicInfo[0];
     }
 
+    /**
+     * Retrieve the set of authentication templates configured from file system in JSON format.
+     *
+     * @return Authentication templates.
+     */
+    public abstract String getAuthenticationTemplatesJSON();
```
madurangasiriwardena: Better to avoid adding default method, as this will break the API

components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/ApplicationManagementService.java
```diff
@@ -331,6 +332,16 @@ public abstract boolean isExistingApplicationTemplate(String templateName, Strin
     public abstract List<SpTemplate> getAllApplicationTemplateInfo(String tenantDomain)
             throws IdentityApplicationManagementException;
 
+    /**
+     * Get inbound authenticator configurations
```
mefarazath: It would be better if we could give an idea of the key and values of the map. For example, what is the key returned in this map?

PR #2557, opened 2019-11-19
Fix umlauts characters are present in the service provider description the service provider export is not working as expected

DMHP:

Fix umlauts characters are present in the service provider description the service provider export is not working as expected

Immediately

### Follow up actions

None


DMHP:
Fix https://github.com/wso2/product-is/issues/6847

PR #2588, opened 2019-11-29
Handle Unique Constraint Violations With Concurrent Requests

tharindu-bandara:
- An application creation, role creation and updating user list of roles scenarios modified with this pull request, are changed to handle the unique constraint errors thrown upon already existing entities, especially when concurrent requests were made.


tharindu-bandara:
Needs to be fixed in a separate PR, with conflicts properly resolved.

PR #2590, opened 2019-12-02
Improvements to Application Management service

mefarazath:
* Improve performance and reduce complexity of createApplication() method. 

This is done by avoiding a GET after create and returning only the created resourceId in the response.

* Rename loginUrl attribute to accessUrl
The reasoning behind the change was that there could be registered applications that do not need to login but simply need a URL to access it (bookmark type applications)


components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/ApplicationResourceManager.java
```diff
@@ -44,7 +44,7 @@ ApplicationBasicInfo getApplicationBasicInfoByResourceId(String resourceId, Stri
      * @return unique application resource id of the application
      * @throws IdentityApplicationManagementException
      */
-    ServiceProvider createApplication(ServiceProvider application, String tenantDomain, String username)
+    String createApplication(ServiceProvider application, String tenantDomain, String username)
```
mefarazath: This API in previous versions of IS 5.9.0 and before was void type. We decided to return ServiceProvider object in 5.10.0 initial milestones. But returning the service object requires an additional GET which is expensive and increases complexity of having to rolling back if GET fails even after create is successful.

PR #2609, opened 2019-12-06
Validate inbound authentication keys before application create/update

emswbandara:
Currently we do not validate whether the inbound authentication key is already existing for another application when creating certain inbound authentication types. E.g. Passive-STS and WS-Trust .
As a result it will produce inconsistent results when using those applications. Purpose of this PR is to add the validation part before creating/updating applications.


components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/ApplicationMgtValidator.java
```diff
@@ -98,6 +105,69 @@ private void validateDiscoverabilityConfigs(List<String> validationErrors,
...
+            if (log.isDebugEnabled()) {
+                log.debug("Cannot find application name for the inbound auth key: " + inboundConfig
+                        .getInboundAuthKey() + " of inbound auth type: " + inboundConfig.getInboundAuthType());
+            }
+            return;
+        }
+        ServiceProvider existingApp = applicationDAO.getApplication(existingAppName, tenantDomain);
+
+        if (existingApp != null && existingApp.getApplicationID() != appId) {
+            String msg = "Inbound key: '" + inboundConfig.getInboundAuthKey() + "' is already configured for the" +
```
mefarazath: let's add the type here.

PR #2610, opened 2019-12-06
Update exception in userstore config admin service

madurangasiriwardena:
`IdentityUserStoreClientException` is added extending `IdentityUserStoreMgtException` to identify the client exception from the OSGi service. However if the admin service does not declare `IdentityUserStoreClientException` is thrown from the service methods, stub will not contain the `IdentityUserStoreClientException` and will fail to throw it from the stub method invocation. Instead, it will throw an AxisFault which is not an ideal option.


components/user-store/org.wso2.carbon.identity.user.store.configuration/src/main/java/org/wso2/carbon/identity/user/store/configuration/UserStoreConfigAdminService.java
```diff
@@ -187,6 +197,8 @@ public void editUserStoreWithDomainName(String previousDomainName, UserStoreDTO
         } catch (UserStoreException e) {
             String errorMessage = e.getMessage();
             throw new IdentityUserStoreMgtException(errorMessage);
+        } catch (IdentityUserStoreClientException e) {
+            throw throwIdentityUserStoreMgtException(e, "Error while updating the userstore.");
```
mefarazath: How about renaming throwIdentityUserStoreMgtException -> buildIdentityUserStoreMgtException?

PR #2614, opened 2019-12-09
Fix performance issue when filtering users.

hasinidilanka:
**Proposed Changes in PR**

Introduce a new configuration in identity.xml as follow to identify whether to show the display name claim (http://wso2.org/claims/displayName) when filtering users.
```
 <UserFiltering> 
        <ShowDisplayName>false</ShowDisplayName>
</UserFiltering>
```

Search for the display name of users only if the configuration is set to true.

The default value of the config is false.

Resolve - https://github.com/wso2/product-is/issues/6937

### DIFF
--- a/components/user-mgt/org.wso2.carbon.user.mgt/src/main/java/org/wso2/carbon/user/mgt/UserRealmProxy.java
+++ b/components/user-mgt/org.wso2.carbon.user.mgt/src/main/java/org/wso2/carbon/user/mgt/UserRealmProxy.java
@@ -25,6 +25,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.wso2.carbon.CarbonConstants;
 import org.wso2.carbon.context.CarbonContext;
+import org.wso2.carbon.identity.base.IdentityConstants;
 import org.wso2.carbon.identity.core.util.IdentityUtil;
 import org.wso2.carbon.registry.api.Registry;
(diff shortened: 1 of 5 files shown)


components/user-mgt/org.wso2.carbon.user.mgt/src/main/java/org/wso2/carbon/user/mgt/UserRealmProxy.java
```diff
@@ -117,15 +118,24 @@ public UserRealmProxy(UserRealm userRealm) {
...
                     flaggedNames[i].setItemName(user);
                     //retrieving the displayName
-                    String displayName = realm.getUserStoreManager().getUserClaimValue(user, DISAPLAY_NAME_CLAIM, null);
-                    int index = user.indexOf(UserCoreConstants.DOMAIN_SEPARATOR);
-                    if (index > 0) {
-                        if (displayName != null) {
+                    // Check whether to use the display name claim when filtering users.
+                    String showDisplayName = IdentityUtil.getProperty(IdentityConstants.SHOW_DISPLAY_NAME);
+                    boolean isShowDisplayNameEnabled = false;
+                    if (StringUtils.isNotBlank(showDisplayName)) {
```
DMHP: No need to check for null. If this value is null or empty the value will be false by default.

hasinidilanka: Fixed.

PR #2616, opened 2019-12-09
Add API specific scopes/permissions to identity.xml

mefarazath:


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml
```diff
@@ -1166,6 +1166,13 @@
             <Scopes>internal_identity_mgt_delete</Scopes>
         </Resource>
         <Resource context="(.*)/api/server/v1/keystores/certs/public(.*)" secured="false" http-method="GET"/>
+        <Resource context="(.*)/api/server/v1/applications(.*)" secured="true" http-method="all">
+            <Permissions>/permission/admin/manage/identity/applicationmgt</Permissions>
+            <Scopes>internal_application_mgt_view</Scopes>
```
madurangasiriwardena: Don't we need to add separate permissions to each http method such as internal_application_mgt_delete for delete operation and so on.

mefarazath: We can make the scopes required for each action. But we cannot always map it to the HTTP method. For DELETE we can. 

But when it comes to PUT it could be either for an update or create. So we might have to have both
internal_application_mgt_create and internal_application_mgt_update. 

I followed the pattern used in user APIs which did not have any granular scopes based on the API action. 

If we require granular permissions for SP API actions, will update the PR.

mefarazath: On second thoughts, I think we can group the scopes in following manner

GET -  internal_application_mgt_view
PUT, POST, PATCH - internal_application_mgt_create internal_application_mgt_update
DELETE - internal_application_mgt_delete

Updated the PR with the changes.

PR #2634, opened 2019-12-18
Remove validation making imageURL required for discoverable apps

darshanasbg:
$subject


mefarazath:
Any specific reason why we decide to remove the validation?

darshanasbg:
> Any specific reason why we decide to remove the validation?

As per the discussion, https://github.com/wso2/product-is/issues/7095, it was identified that, its a overhead to admin, to configure imageURL to make application discoverable as we don't support uploading images in IS itself as of now..

From the UI perspective imageURL shouldn't be a mandatory field that identity admin should fill. But API side it can be a mandatory one.. To achieve that, we should have the capability like following,

1. Admin can add the link of a `imageURL` or upload the SP image from the IS itself.
2. If admin does not do that, UI should generate a default image and use that as the `imageURL`.
3. So the create or the update API request of the SP will always have `imageURL`.

As we are missing the ability to upload an SP image as of now, and we don't have the capability add default image when registering the SP (as we are not trying to add much more features to management console), removed the validation with this PR..

We will have the the flow as above, when we have the `Admin Portal`

PR #2659, opened 2020-01-03
Username filter contains illegal character error

thushaaanthan:
issue: wso2/product-is#7174

When we have users with user name as "A'KELAH" (with Apostrophe). When the admin user tried to search users via the management console by user name an error is thrown.


components/identity-core/org.wso2.carbon.identity.core.ui/src/main/resources/web/identity/validation/js/identity-validate.js
```diff
@@ -75,6 +75,9 @@ function getPattern(pattern) {
         case "fragment-free-url":
             regex = /^(([^:/?#]+):)?(([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?([^#])?/;
             break;
+        case "user-search-exists":
```
mefarazath: user-search-exists doesn't seem to convey the intended purpose

PR #2713, opened 2020-01-29
Revert changes done in #2655

pulasthi7:
Reverting the changes based on the offline discussion. This approach will not fix the initial problem reported. Instead, we decided to redirect users to an endpoint at application level to terminate the session there and initiate a logout request to terminate the session at identity server also.


senthalan:
Linking the reverted PR for the reference https://github.com/wso2/carbon-identity-framework/pull/2655/files

PR #2714, opened 2020-01-29
Continue IdentityUserIdResolverListener operation when the user list is empty

madurangasiriwardena:


components/identity-mgt/org.wso2.carbon.identity.mgt/src/main/java/org/wso2/carbon/identity/mgt/listener/IdentityUserIdResolverListener.java
```diff
@@ -531,7 +531,16 @@ public boolean doPreAddRole(String roleName, String[] userList, Permission[] per
...
+        String[] userIDs;
+        if (ArrayUtils.isNotEmpty(userList)) {
+            userIDs = getUserIdsFromUserNames(userList, (AbstractUserStoreManager) userStoreManager);
+
+            if (ArrayUtils.isEmpty(userIDs)) {
+                return true;
+            }
+        } else {
+            userIDs = new String[0];
+        }
```
tharindu-bandara: Shall we use a turnery operator here to improve readability and reduce empty check on the userIDs?

`userIDs = ArrayUtils.isNotEmpty(userList) ? getUserIdsFromUserNames(userList, (AbstractUserStoreManager) userStoreManager) : new String[0];`

madurangasiriwardena: We need to return only if the `userList` is not empty and `userIDs` array returned from the `getUserIdsFromUserNames` method invocation is empty. So the above mentioned approach will not work.

components/identity-mgt/org.wso2.carbon.identity.mgt/src/main/java/org/wso2/carbon/identity/mgt/listener/IdentityUserIdResolverListener.java
```diff
@@ -531,7 +531,16 @@ public boolean doPreAddRole(String roleName, String[] userList, Permission[] per
...
 
         if (ArrayUtils.isEmpty(userIDs)) {
```
tharindu-bandara: We need to remove this check as this will override the fix.

components/identity-mgt/org.wso2.carbon.identity.mgt/src/main/java/org/wso2/carbon/identity/mgt/listener/IdentityUserIdResolverListener.java
```diff
@@ -531,7 +531,16 @@ public boolean doPreAddRole(String roleName, String[] userList, Permission[] per
...
             return true;
```
tharindu-bandara: We need to remove this check as this will override the fix.

PR #2732, opened 2020-02-07
Fix set sameSite value for commonAuth cookie

thushaaanthan:
Fix for:  https://github.com/wso2/product-is/issues/7477


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/util/FrameworkUtils.java
```diff
@@ -30,6 +30,7 @@
 import org.wso2.carbon.CarbonConstants;
 import org.wso2.carbon.context.CarbonContext;
 import org.wso2.carbon.context.PrivilegedCarbonContext;
+import org.wso2.carbon.core.SameSiteCookie;
```
malithie: Did we name it as SameSiteCookie ?
IIRC we discussed to make this extended cookie class capable of handling any explicit attributes if needed.
If so, it should have a different naming

malithie: So the extended cookie seems to be 'ServletCookie' and SameSiteCookie just holds same site modes right.
Then +1

PR #2733, opened 2020-02-10
[FIX] Encoding password regex

somindatommy:
Resolves wso2/product-is#7505

The issue is due to not encoding `>` character properly. This PR encodes `>` to `&gt;`.

## Related PRs
Support Fix : https://github.com/wso2-support/carbon-identity-framework/pull/950

### DIFF
--- a/components/user-mgt/org.wso2.carbon.user.mgt.ui/src/main/resources/web/user/add-step1.jsp
+++ b/components/user-mgt/org.wso2.carbon.user.mgt.ui/src/main/resources/web/user/add-step1.jsp
@@ -59,6 +59,9 @@
     boolean isAskPasswordEnabled = false;
     boolean isUserOnBoardingEnabled = false;
 
+    String GREATER_THAN_CHARACTER = ">";
+    String GREATER_THAN_ENCODING_STRING = "&gt;";
+
     String BUNDLE = "org.wso2.carbon.userstore.ui.i18n.Resources";
     ResourceBundle resourceBundle = ResourceBundle.getBundle(BUNDLE, request.getLocale());
 
@@ -343,9 +346,11 @@
             <form method="post" action="add-finish-ajaxprocessor.jsp" name="dataForm" onsubmit="return doValidation();">
 
                 <input type="hidden" id="pwd_primary_null" name="pwd_primary_null"
-                       value=<%=Encode.forHtmlAttribute(userRealmInfo.getPrimaryUserStoreInfo().getPasswordRegEx())%>>
+                       value=<%=Encode.forHtmlAttribute(userRealmInfo.getPrimaryUserStoreInfo().getPasswordRegEx())
+                       .replace(GREATER_THAN_CHARACTER, GREATER_THAN_ENCODING_STRING)%>>
                 <input type="hidden" id="usr_primary_null" name="usr_primary_null"
-                       value=<%=Encode.forHtmlAttribute(userRealmInfo.getPrimaryUserStoreInfo().getUserNameRegEx())%>>
+                       value=<%=Encode.forHtmlAttribute(userRealmInfo.getPrimaryUserStoreInfo().getUserNameRegEx())
+                       .replace(GREATER_THAN_CHARACTER, GREATER_THAN_ENCODING_STRING)%>>
 
                 <%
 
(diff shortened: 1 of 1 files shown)


components/user-mgt/org.wso2.carbon.user.mgt.ui/src/main/resources/web/user/add-step1.jsp
```diff
@@ -360,11 +365,11 @@
                 <input type="hidden"
                        id="pwd_<%=Encode.forHtmlAttribute(allUserStoreInfo[i].getDomainName().toUpperCase())%>"
                        name="pwd_<%=Encode.forHtmlAttribute(allUserStoreInfo[i].getDomainName().toUpperCase())%>"
-                       value=<%=Encode.forHtmlAttribute(pwdRegEx)%>>
+                       value=<%=Encode.forHtmlAttribute(pwdRegEx).replace(REATER_THAN_CHARACTER,GREATER_THAN_ENCODING_STRING)%>>
```
vihanga-liyanage: Formatting issues.

somindatommy: Fixed

components/user-mgt/org.wso2.carbon.user.mgt.ui/src/main/resources/web/user/add-step1.jsp
```diff
@@ -59,6 +59,9 @@
     boolean isAskPasswordEnabled = false;
     boolean isUserOnBoardingEnabled = false;
 
+    String GREATER_THAN_CHARACTER = ">";
```
tharindu-bandara: @somindatommy 

This is fixing only the character `>`. Is that intentional? 

Means the only character is not encoded is `>`

somindatommy: @tharindu-bandara yes. Other characters such as `"`,`'`,`&` and `<` are encoded properly. Issue arises as `>` is not encoded.

tharindu-bandara: In the API docs, it says,

> The encoding of the greater-than sign (>) is not required for attributes.

I'm wondering what could be the reason behind this. 

Maybe the encoding that we do here does not meet the expectation? For instance, [`Encode.forHtmlContent`](https://www2.owasp.org/owasp-java-encoder/encoder/apidocs/org/owasp/encoder/Encode.html#forHtmlContent-java.lang.String-) does encode the character `>`.


[1] [https://www2.owasp.org/owasp-java-encoder/encoder/apidocs/org/owasp/encoder/Encode.html#forHtmlAttribute-java.lang.String-](https://www2.owasp.org/owasp-java-encoder/encoder/apidocs/org/owasp/encoder/Encode.html#forHtmlAttribute-java.lang.String-)

somindatommy: In the API docs, it says the following about `Encode.forHtmlContent` [1]

> It does not escape quotation characters and is thus unsafe for use with HTML attributes. Use either forHtml or forHtmlAttribute for those methods.

Hence `Encode.forHtmlAttribute` is used. To prevent HTML injection, encoding `<` as `&lt;` is enough. Therefore encoding `>` is not mandatory.

When `>` is not encoded, it is considered as a closing tag causing rendering errors in the web page. That's why encoding is necessary. 

[1] - https://www2.owasp.org/owasp-java-encoder/encoder/apidocs/org/owasp/encoder/Encode.html#forHtmlContent-java.lang.String-

tharindu-bandara: @somindatommy Thanks for the findings and the explanation!

Then we can use the [encoding for html](https://www2.owasp.org/owasp-java-encoder/encoder/apidocs/org/owasp/encoder/Encode.html#forHtml-java.lang.String-) ? Seems like this addressed the greater than the character as well.

somindatommy: Fixed: 225deac693d3998c3919586cdf7b829647e41a9c

PR #2770, opened 2020-02-19
Fix for confirmation email resending issue

AnuradhaSK:

Resolves https://github.com/wso2/product-is/issues/7643


features/authentication-framework/org.wso2.carbon.identity.application.authentication.framework.server.feature/pom.xml
```diff
@@ -46,6 +46,32 @@
...
+                        <id>copy</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>copy</goal>
+                        </goals>
+                        <configuration>
+                            <artifactItems>
+                                <artifactItem>
+                                    <groupId>org.wso2.carbon.identity.framework</groupId>
+                                    <artifactId>org.wso2.carbon.identity.application.authentication.endpoint.util</artifactId>
```
senthalan: Why we copy torg.wso2.carbon.identity.application.authentication.endpoint.util to the runtimes, this jar is already available it the plugins. Duplicating jars in two places is not recommended. If there is an issue with getting class from this jar in webapp level, we need to fix this other than coping it.

madurangasiriwardena: This is removed from the plugins and added to cxf runtime, mainly because this should only be needed by the webapps, but not in the OSGi runtime.

senthalan: okay. sorry for the confusion. In this case, can we add to a different runtime environment and include that environment in the needed web apps. If we add to cxf3, this jar will be loaded for all the CXF web apps in the start-up. But I think only some apps use this

AnuradhaSK: If org.wso2.carbon.identity.application.authentication.endpoint.util  component is copied only to the 'authentication endpoint' web app libs, the following error is logged when the server gets up,
Error configuring application listener of class [org.wso2.carbon.identity.application.authentication.endpoint.util.listener.AuthenticationEndpointContextListener] java.lang.ClassNotFoundException: org.wso2.carbon.identity.application.authentication.endpoint.util.listener.AuthenticationEndpointContextListener

PR #2781, opened 2020-02-24
Introduce onUserStorePreDisable() handler in UserStoreConfigListener

gayashanbc:
Related https://github.com/wso2/product-is/issues/5227


components/user-store/org.wso2.carbon.identity.user.store.configuration/src/main/java/org/wso2/carbon/identity/user/store/configuration/UserStoreConfigAdminService.java
```diff
@@ -324,6 +325,15 @@ public void changeUserStoreState(String domain, Boolean isDisable) throws Identi
             TransformerConfigurationException {
 
         validateDomain(domain, isDisable);
+
+        try {
+            if (isDisable) {
+                triggerListenersOnUserStorePreDisable(domain);
```
IsuraD: Shall we pass the modified status? Then we can use this for enabling as well

PR #2807, opened 2020-03-03
Use String Builders instead of Strings to enhance performance 

AnuradhaSK:
Change String objects defined in IdentityProviderManager.java file to string builders in order to enhance performance.


thanujalk:
@AnuradhaSK Do we really need this improvement ? https://stackoverflow.com/questions/1532461/stringbuilder-vs-string-concatenation-in-tostring-in-java

AnuradhaSK:
The requirement of this improvement will be analysed again later via a product performance tests. Therefore, Closing this PR at the moment.

PR #2826, opened 2020-03-16
Modify the IDP listing by skipping resident IDP

GANGANI:
Resolves: https://github.com/wso2/product-is/issues/7886

### DIFF
--- a/components/idp-mgt/org.wso2.carbon.idp.mgt/src/main/java/org/wso2/carbon/idp/mgt/dao/IdPManagementDAO.java
+++ b/components/idp-mgt/org.wso2.carbon.idp.mgt/src/main/java/org/wso2/carbon/idp/mgt/dao/IdPManagementDAO.java
@@ -288,6 +288,8 @@ public class IdPManagementDAO {
                                          String sortOrder, String sortBy)
             throws IdentityProviderManagementServerException, IdentityProviderManagementClientException {
 
+        // Add 1 to offset as the Resident IDP is the 1st entry in the IDP table.
+        offset = offset +1;
         FilterQueryBuilder filterQueryBuilder = new FilterQueryBuilder();
         appendFilterQuery(expressionNode, filterQueryBuilder);
         String sortedOrder = sortBy + " " + sortOrder;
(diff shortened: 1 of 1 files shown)


components/idp-mgt/org.wso2.carbon.idp.mgt/src/main/java/org/wso2/carbon/idp/mgt/dao/IdPManagementDAO.java
```diff
@@ -288,6 +288,8 @@
                                          String sortOrder, String sortBy)
             throws IdentityProviderManagementServerException, IdentityProviderManagementClientException {
 
+        // Add 1 to offset as the Resident IDP is the 1st entry in the IDP table.
+        offset = offset + 1;
```
emswbandara: since it is not safe to assume that the first entry of the search result is the local IDP specially in tenant case, let's update the SQL query to list IDPs to filter out LOCAL IDP. This should be similar to the count SQL query.

[1] https://github.com/wso2/carbon-identity-framework/blob/master/components/idp-mgt/org.wso2.carbon.idp.mgt/src/main/java/org/wso2/carbon/idp/mgt/util/IdPManagementConstants.java#L107

GANGANI: Local IDP is not listing with other IDPs. But as the local IDP is in the DB, it affects when listing IDP using the limit and the offset.

emswbandara: i think instead of handling from the code to filter out resident IDP from the IDP list, we need to handle it from the SQL query itself. see https://github.com/wso2/carbon-identity-framework/pull/2826/files#diff-aef45437564b8a5e30cf2197682d6fa6L448-L451

emswbandara:
Closing since issue has been fixed with https://github.com/wso2/carbon-identity-framework/pull/2837

PR #2831, opened 2020-03-18
Fix issue of session termination

GANGANI:
### Description
Fix the issue of session termination on password change. Get the user's unique ID to terminate the session.

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
@@ -81,16 +81,11 @@ public class UserSessionManagementServiceImpl implements UserSessionManagementSe
         }
     }
 
+    // This method invokes only for the local users to resolve the user's unique user id from username.
     private List<String> getSessionsOfUser(String username, String userStoreDomain, String tenantDomain) throws
             UserSessionException {
 
-        String userId;
-        if (StringUtils.isEmpty(userStoreDomain) || FrameworkConstants.FEDERATED_IDP_NAME.equals(userStoreDomain)){
-            userId = UserSessionStore.getInstance().getUserId(username, getTenantId(tenantDomain),
-                    userStoreDomain);
-        } else {
-            userId = FrameworkUtils.resolveUserIdFromUsername(getTenantId(tenantDomain), userStoreDomain, username);
-        }
+        String userId = FrameworkUtils.resolveUserIdFromUsername(getTenantId(tenantDomain), userStoreDomain, username);
         return UserSessionStore.getInstance().getSessionId(userId);
     }
 
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
@@ -85,6 +85,9 @@ public class UserSessionManagementServiceImpl implements UserSessionManagementSe
 
         String userId = UserSessionStore.getInstance().getUserId(username, getTenantId(tenantDomain),
                 userStoreDomain);
+        if (userId == null) {
+            userId = FrameworkUtils.resolveUserIdFromUsername(getTenantId(tenantDomain), userStoreDomain, username);
+        }
         return UserSessionStore.getInstance().getSessionId(userId);
     }
 
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
@@ -83,11 +83,7 @@ public class UserSessionManagementServiceImpl implements UserSessionManagementSe
     private List<String> getSessionsOfUser(String username, String userStoreDomain, String tenantDomain) throws
             UserSessionException {
 
-        String userId = UserSessionStore.getInstance().getUserId(username, getTenantId(tenantDomain),
-                userStoreDomain);
-        if (userId == null) {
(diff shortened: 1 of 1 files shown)


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
```diff
@@ -85,6 +85,9 @@ private void validate(String username, String userStoreDomain, String tenantDoma
 
         String userId = UserSessionStore.getInstance().getUserId(username, getTenantId(tenantDomain),
```
janakamarasena: What happens when we do this call?
UserSessionStore.getInstance().getUserId(username, getTenantId(tenantDomain), userStoreDomain);

janakamarasena: @pamodaaw can we replace this with the below logic? will this method be called for federated users by anychance?

GANGANI: > What happens when we do this call?
> UserSessionStore.getInstance().getUserId(username, getTenantId(tenantDomain), userStoreDomain);

That getUserID method call is to return the user Id of a user from the database table of IDN_AUTH_USER. Therefore it returns null.

pamodaaw: Yes, we can remove the user-id validation from the IDN_AUTH_USER table since we add only the federated user IDs to this table.

malithie:
Didn't get the problem here.
Can we please describe the problem and the approach taken to fix it

GANGANI:
> Didn't get the problem here.
> Can we please describe the problem and the approach taken to fix it

We stopped populating the IDN_AUTH_USER table for local users as we have the unique ID capability now. So now when we check from the IDN_AUTH_USER table for a local user ID it is not there. Therefore we need to check for Local user IDs in a new way and for federated user IDs in the IDN_AUTH_USER table

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
```diff
@@ -83,8 +84,13 @@ private void validate(String username, String userStoreDomain, String tenantDoma
     private List<String> getSessionsOfUser(String username, String userStoreDomain, String tenantDomain) throws
             UserSessionException {
 
-        String userId = UserSessionStore.getInstance().getUserId(username, getTenantId(tenantDomain),
-                userStoreDomain);
+        String userId;
+        if (StringUtils.isEmpty(userStoreDomain) || FrameworkConstants.FEDERATED_IDP_NAME.equals(userStoreDomain)){
+            userId = UserSessionStore.getInstance().getUserId(username, getTenantId(tenantDomain),
+                    userStoreDomain);
```
malithie: Looking at the triggering point of this call [1], it doesn't look like a user store domain can be "FEDERATED".
Did we verify this
[1] https://github.com/wso2/carbon-identity-framework/blob/master/components/identity-mgt/org.wso2.carbon.identity.mgt/src/main/java/org/wso2/carbon/identity/mgt/listener/UserSessionTerminationListener.java#L134

malithie: Basically this method would never be invoked for a federated it seems, unless we provide an API and consume the OSGI service for a federated user case.
@pamodaaw , did we do such ?

pamodaaw: We don't have such API for federated users which requires the OSGi service. When storing the user details in the IDN_AUTH_USER table, if the userstore domain is invalid, we assume that the user is a federated user and in the entry, the user store domain is marked as FEDERATED.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
```diff
@@ -80,11 +81,11 @@ private void validate(String username, String userStoreDomain, String tenantDoma
         }
     }
 
+    // This method invokes only for the local users to resolve the user's unique user id from username.
```
pamodaaw: Shall we add a more descriptive comment for the function? Since the function is to retrieve the sessions, move the user-id related comment near the logic resolving the user-id

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
```diff
@@ -80,11 +81,11 @@ private void validate(String username, String userStoreDomain, String tenantDoma
...
     private List<String> getSessionsOfUser(String username, String userStoreDomain, String tenantDomain) throws
             UserSessionException {
 
-        String userId = UserSessionStore.getInstance().getUserId(username, getTenantId(tenantDomain),
-                userStoreDomain);
+        String userId = FrameworkUtils.resolveUserIdFromUsername(getTenantId(tenantDomain), userStoreDomain, username);
         return UserSessionStore.getInstance().getSessionId(userId);
```
malithie: How this exactly should be fixed is,
- the legacy API method terminateSessionsOfUser(String username, String userStoreDomain, String tenantDomain) should consume the new API method terminateSessionsByUserId(String userId)
- therefore, within terminateSessionsOfUser(String username, String userStoreDomain, String tenantDomain) you will resolve the user id invoking FrameworkUtils.resolveUserIdFromUsername(getTenantId(tenantDomain), userStoreDomain, username) and then invoke terminateSessionsByUserId(String userId)
- So, eventually when federated user case comes into picture FrameworkUtils.resolveUserIdFromUsername(getTenantId(tenantDomain), userStoreDomain, username should be able to resolve the user id for such users.
- This make the private method which you have used redundant. so you can remove it actually

madurangasiriwardena:
Fix will be added with https://github.com/wso2-support/carbon-identity-framework/pull/1045. Hence closing the PR.

PR #2828, opened 2020-03-17
Add tenant-qualified URL support for endpoints

ChamathNS:
### Proposed changes in this pull request

Add a service for resolving URLs and tenant-qualifying endpoints.
Tenant-qualify SAML metadata. 
fixes: https://github.com/wso2/product-is/issues/7900

PR #2838, opened 2020-03-25
Add support for including tenant domain as a parameter in URLResolverService

ChamathNS:

Overload the current method implementations to accommodate tenant domain as a parameter in URLResolverService. 

Improvement for: https://github.com/wso2/carbon-identity-framework/pull/2828
Related to: https://github.com/wso2/product-is/issues/7902

### DIFF
--- a/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultURLResolverService.java
+++ b/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultURLResolverService.java
@@ -40,6 +40,22 @@ import static org.wso2.carbon.identity.core.util.IdentityTenantUtil.isTenantQual
 public class DefaultURLResolverService implements URLResolverService {
 
     @Override
+    public String resolveUrl(String url, boolean addProxyContextPath, boolean addWebContextRoot,
+                             Map<String, Object> properties) throws URLResolverException {
+
+        return resolveUrl(url, null, addProxyContextPath, addWebContextRoot, false, false, properties);
+    }
+
+    @Override
+    public String resolveUrl(String url, boolean addProxyContextPath, boolean addWebContextRoot,
+                             boolean addTenantQueryParamInLegacyMode, boolean addTenantPathParamInLegacyMode,
+                             Map<String, Object> properties) throws URLResolverException {
+
+        return resolveUrl(url, null, addProxyContextPath, addWebContextRoot, addTenantQueryParamInLegacyMode,
+                addTenantPathParamInLegacyMode, properties);
+    }
+
+    @Override
     public String resolveUrl(String url, String tenantDomain, boolean addProxyContextPath, boolean addWebContextRoot,
                              boolean addTenantQueryParamInLegacyMode, boolean addTenantPathParamInLegacyMode,
                              Map<String, Object> properties) throws URLResolverException {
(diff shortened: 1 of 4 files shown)


components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultURLResolverService.java
```diff
@@ -40,7 +40,7 @@
 public class DefaultURLResolverService implements URLResolverService {
 
     @Override
-    public String resolveUrl(String url, boolean addProxyContextPath, boolean addWebContextRoot,
+    public String resolveUrl(String url, String tenantDomain, boolean addProxyContextPath, boolean addWebContextRoot,
```
malithie: Shall we have a new API to accept tenant domain. 
Wondering if there is s real necessity of it as well.

In addition, we should have an API without legacy mode params as well. So new consumers don't need to worry about populating such.

ChamathNS: We have a new API for this at the IdentityUtil.class 
There's an another API available at IdentityUtil without legacy mode params.

There is a need for this because there are instances where tenantDomain is directly available as a attribute (for an instance, while tenant creation). So then, with this change, we don't need to meddle with the ThreadLocal or the CarbonContext to get the tenant domain.

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultURLResolverService.java
```diff
@@ -135,11 +168,13 @@ private void appendProxyContextPath(StringBuilder serverUrl) {
...
+    private void appendTenantAsPathParam(StringBuilder serverUrl, String tenantDomain) {
 
-        String tenantDomain = IdentityTenantUtil.getTenantDomainFromContext();
         if (StringUtils.isBlank(tenantDomain)) {
-            tenantDomain = PrivilegedCarbonContext.getThreadLocalCarbonContext().getTenantDomain();
+            tenantDomain = IdentityTenantUtil.getTenantDomainFromContext();
+            if (StringUtils.isBlank(tenantDomain)) {
+                tenantDomain = PrivilegedCarbonContext.getThreadLocalCarbonContext().getTenantDomain();
+            }
         }
```
malithie: Shall we move this tenant resolving logic to a single method and reuse ?

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultURLResolverService.java
malithie: But the API should be clean.
So looks like this is addressed

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultURLResolverService.java
```diff
@@ -166,16 +193,29 @@ private void appendTenantPathParamInLegacyMode(StringBuilder serverUrl) {
...
     }
 
-    private void appendTenantQueryParamInLegacyMode(StringBuilder serverUrl) {
+    private void appendTenantQueryParamInLegacyMode(StringBuilder serverUrl, String tenantDomain) {
 
-        String tenantDomain = IdentityTenantUtil.getTenantDomainFromContext();
+        if (StringUtils.isBlank(tenantDomain)) {
+            tenantDomain = IdentityTenantUtil.getTenantDomainFromContext();
+        }
```
malithie: reason to not use the resolving method.
Looks like some checks/steps missed here

PR #2848, opened 2020-04-01
Remove redundant usage of 'callbackURL' query param

gayashanbc:
### Description
Goal of this is PR to remove redundant usage of the 'callbackURL' query parameter in the XACML Policy Editor.


mefarazath:
Can you explain what's the issue we are trying to fix here? May be linking the relevant git issue would do.

PR #2851, opened 2020-04-01
Implement serverURL model

ChamathNS:

improvement for https://github.com/wso2/carbon-identity-framework/pull/2828


ChamathNS:
@malithie @pulasthi7 please review.

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServerURL.java
```diff
@@ -0,0 +1,411 @@
...
+    private String protocol;
+    private String hostName;
+    private Integer port;
+    private String proxyContext;
+    private String webContext;
+    private Boolean addProxyContext;
+    private Boolean addWebContext;
+    private String urlContext;
+    private Boolean appendTenantAsQueryParam;
+    private Boolean appendTenantAsPathParam;
```
malithie, line 52: I don't think these properties are needed. Those are just builder properties.
We need just the basic components of the URL like host, port, protocol, path, query, fragement, parameters. May be you can have an internal URL object

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServerURL.java
```diff
@@ -0,0 +1,411 @@
...
+
+import java.net.SocketException;
+import java.util.ArrayList;
+
+import static org.wso2.carbon.identity.core.util.IdentityTenantUtil.isTenantQualifiedUrlsEnabled;
+
+/**
+ * Server URL representation.
+ */
+public class ServerURL {
```
malithie, line 41: What if we say this as ServiceURL; This actually reflects service endpoints.
@pulasthi7 @mefarazath @omindu WDYT?

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServerURL.java
```diff
@@ -0,0 +1,411 @@
...
+
+    public Boolean getAppendTenantAsQueryParam() {
+
+        return appendTenantAsQueryParam;
+    }
+
+    public Boolean getAppendTenantAsPathParam() {
+
+        return appendTenantAsPathParam;
+    }
```
malithie, line 120: Remove  getters related to unnecessary operational properties

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServerURL.java
```diff
@@ -0,0 +1,411 @@
...
+
+    public ArrayList<String> getParameters() {
+
+        return parameters;
+    }
```
malithie, line 125: Don't return the list which is mutable. Rather return as properties, immutable list, or better to provide the capability to get a specific parameter given a key

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServerURL.java
```diff
@@ -0,0 +1,411 @@
...
+
+    public ArrayList<String> getFragments() {
+
+        return fragments;
+    }
+
+    private String fetchProtocol() {
+
+        return CarbonUtils.getManagementTransport();
+    }
```
malithie, line 135: Let's not use carbon utils and fetch. This object should be created with all necessary fields populated. Don't invoke other classes utils from this POJO

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServerURL.java
```diff
@@ -0,0 +1,411 @@
...
+            this.fragments = fragments;
+            return this;
+        }
+
+        public ServerURL build() {
+
+            return new ServerURL(this);
+        }
+    }
+
```
malithie, line 410: We need a slightly deviated version of this. Builder is the one who should process and get the POJO returned

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServerURL.java
```diff
@@ -0,0 +1,411 @@
...
+            return this;
+        }
+
+        public ServerURLBuilder appendTenantAsPathParam() {
+
+            this.appendTenantAsPathParam = true;
+            return this;
+        }
+
+        public ServerURLBuilder parameters(ArrayList<String> parameters) {
```
mefarazath, line 393: Also would be better to have a parameter() method that allows adding params one by one.
Also shouldn't the input be a map?

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,212 @@
...
+
+import static org.wso2.carbon.identity.core.util.IdentityTenantUtil.isTenantQualifiedUrlsEnabled;
+
+/**
+ * Implementation for {@link ServiceURLBuilder}.
+ * Builder for {@link ServiceURL} instances.
+ */
+public class DefaultServiceURLBuilder implements ServiceURLBuilder {
+
+    private Boolean addProxyContext;
```
malithie, line 44: Why can't we have the primitive type

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,212 @@
...
+    private Boolean addWebContext;
+    private String urlContext;
+    private Boolean appendTenantAsQueryParam;
+    private Boolean appendTenantAsPathParam;
+    private Map<String, Object> parameters;
```
malithie, line 49: We can String as values right. Even if you take servlet API parameters are string isn't it

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,212 @@
...
+    private Map<String, Object> fragments;
```
malithie, line 50: Same case applies here. I don't think it's necessary to have an object type

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,212 @@
...
+    public ServiceURLBuilder appendTenantAsPathParam() {
+
+        this.appendTenantAsPathParam = true;
+        return this;
+    }
+
+    @Override
+    public ServiceURLBuilder addParameter(String key, Object value) {
+
+        return null;
```
malithie, line 90: This is pending to implement ?

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    @Override
+    public ServiceURL build() throws URLResolverException {
```
malithie, line 94: Shall we change the exception type

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+import org.wso2.carbon.identity.core.model.ServiceURL;
+
+/**
+ * URL Builder service interface.
+ */
+public interface ServiceURLBuilder {
+
+    static ServiceURLBuilder create() {
+
+        return new DefaultServiceURLBuilder();
```
malithie, line 30: Should pick this up from OSGI service listening

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /**
+     * This method is called to add the URL context to the builder.
+     *
+     * @param context URL context.
+     * @return {@link ServiceURLBuilder}.
+     */
+    ServiceURLBuilder addContext(String context);
```
malithie, line 39: I think multiple context paths can be added and you should be building the requestURI

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,212 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.carbon.identity.core;
```
malithie, line 19: This implementation should go to internal package

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    private String resolveUrlFromStandardContext(String urlContext) {
+        /**
+         * TODO
+         * resolve URL context from the standard context.
+         *
+         */
+        return urlContext;
+    }
```
malithie, line 202: We might not need this. context is the path, which we will be using to construct the full URL

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,212 @@
...
+            }
+            resolvedUrlContext = resolveUrlFromStandardContext(urlContext);
+        }
+
+        if (!isTenantQualifiedUrlsEnabled() && appendTenantAsPathParam) {
+            if (StringUtils.isNotBlank(tenantDomain) &&
+                    !MultitenantConstants.SUPER_TENANT_DOMAIN_NAME.equalsIgnoreCase(tenantDomain)) {
+                resolvedUrlStringBuilder.append("/t/").append(tenantDomain);
+            }
+        }
```
malithie, line 129: If not tenant url supported we should give precedence to load the URL from the config

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
ChamathNS: Didn't get what's meant by 'multiple context paths' here. Is it the web context and the proxy context?

malithie: multiple paths.
like addContext(oauth2).addContext(token) should add /oauth2/token

malithie: Maybe it's better to say 'addPath(String... paths)' and provide the capability to add multiple paths with a comma separated string

mefarazath: I think addPath approach is better

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * This method is called when the URL needs to be appended with the tenant domain as a query parameter when
+     * operating in legacy mode.
+     *
+     * @return {@link ServiceURLBuilder}.
+     */
+    ServiceURLBuilder appendTenantAsQueryParam();
```
malithie, line 42: Also let's not make this append right. We can say addTenantQueryParam

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * This method is called when the URL needs to be appended with the tenant domain as a path parameter when
+     * operating in legacy mode.
+     *
+     * @return {@link ServiceURLBuilder}.
+     */
+    ServiceURLBuilder appendTenantAsPathParam();
```
malithie, line 50: Also let's not make this append right. We can say addTenantPathParam

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
```diff
@@ -0,0 +1,174 @@
...
+                serverUrl.append(contextPath.trim());
+            }
+        }
+    }
+
+    private void appendParamsToUri(StringBuilder serverUrl, Map<String, String> parameters) {
+
+        if (MapUtils.isNotEmpty(parameters)) {
+            for (Map.Entry<String, String> entry : parameters.entrySet()) {
+                serverUrl.append("?").append(entry.getKey()).append("=").append(entry.getValue());
```
mefarazath, line 109: We need to do URL encoding here

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
```diff
@@ -0,0 +1,174 @@
...
+        appendContextToUri(absoluteUrl, proxyContextPath);
+
+        String webContextRoot = getWebContext();
+        appendContextToUri(absoluteUrl, webContextRoot);
+
+        String urlContext = getUrlContext();
+        appendContextToUri(absoluteUrl, urlContext);
+
+        Map<String, String> parameters = getParameters();
+        appendParamsToUri(absoluteUrl, parameters);
```
mefarazath, line 147: Shouldn't we handle fragments  here?

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
```diff
@@ -0,0 +1,174 @@
...
+     *
+     * @return The relative URL from the {@link ServiceURL} instance.
+     */
+    public String getRelativeUrl() {
+
+        StringBuilder relativeUrl = new StringBuilder();
+        String urlContext = getUrlContext();
+        appendContextToUri(relativeUrl, urlContext);
+
+        Map<String, String> parameters = getParameters();
```
mefarazath, line 166: Need to handle URL fragements

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * This method is called when the URL needs to be appended with fragments. Fragments should be
+     * entered as <k,v> pairs.
+     *
+     * @param key   Key.
+     * @param value Value.
+     * @return {@link ServiceURLBuilder}.
+     */
+    ServiceURLBuilder addFragments(String key, String value);
```
malithie, line 84: Fragments would not need key value pair I think

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
```diff
@@ -0,0 +1,174 @@
...
+     */
+    public String getAbsoluteUrl() {
+
+        StringBuilder absoluteUrl = new StringBuilder();
+        absoluteUrl.append(getProtocol()).append("://");
+
+        String hostName = getHostName();
+        if (hostName.endsWith("/")) {
+            hostName = hostName.substring(0, hostName.length() - 1);
+        }
```
mefarazath, line 128: Can't we do this within getHostName() method

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
malithie: @mefarazath

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,260 @@
...
+            tenantDomain = PrivilegedCarbonContext.getThreadLocalCarbonContext().getTenantDomain();
+        }
+
+        String resolvedUrlContext = null;
+        StringBuilder resolvedUrlStringBuilder = new StringBuilder();
+        if (isTenantQualifiedUrlsEnabled()) {
+            if (StringUtils.isNotBlank(tenantDomain)) {
+                resolvedUrlStringBuilder.append("/t/").append(tenantDomain);
+            }
+            resolvedUrlContext = resolveUrlFromStandardContext(urlContext);
```
malithie, line 174: we would not have to resolve the context in this mode

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,260 @@
...
+    @Override
+    public ServiceURLBuilder addFragments(String key, String value) {
+
+        this.fragments.put(key, value);
+        return this;
+    }
+
+    @Override
+    public ServiceURL build() throws URLBuilderException {
+
```
malithie, line 148: The very first check we should do here if if tenant url mode is enabled. If not there's a config matched to the context we should return it.
Else we should build

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
```diff
@@ -0,0 +1,174 @@
...
+
+        return webContext;
+    }
+
+    public String getUrlContext() {
+
+        return urlContext;
+    }
+
+    public Map<String, String> getParameters() {
```
malithie, line 84: We should not return the mutable map.
Make map unmodifiable. Also add a method get the value from the parameter key and a method to get just parameter names

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
```diff
@@ -0,0 +1,174 @@
...
+
+        return parameters;
+    }
+
+    public Map<String, String> getFragments() {
+
+        return fragments;
+    }
```
malithie, line 92: Same above applies here

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+
+    /**
+     * Concatenate the protocol, host name, port, proxy context path, web context root and url context to return the
+     * absolute URL.
+     *
+     * @return The absolute URL from the {@link ServiceURL} instance.
+     */
+    public String getAbsoluteUrl() {
```
malithie, line 120: What if we build this at once with the constructor. Or build at first invocation.
There's no way these fields can change later as this class is immutable

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
```diff
@@ -0,0 +1,174 @@
...
+        String urlContext = getUrlContext();
+        appendContextToUri(relativeUrl, urlContext);
+
+        Map<String, String> parameters = getParameters();
+        appendParamsToUri(relativeUrl, parameters);
+
+        if (relativeUrl.toString().endsWith("/")) {
+            relativeUrl.setLength(relativeUrl.length() - 1);
+        }
+        return relativeUrl.toString();
```
malithie, line 172: Same above applies here

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityUtil.java
```diff
@@ -451,12 +453,34 @@ public static String getServerURL(String urlContext, boolean addProxyContextPath
                                       boolean addTenantQueryParamInLegacyMode, boolean addTenantPathParamInLegacyMode)
             throws IdentityRuntimeException {
 
-        URLResolverService urlResolverService = IdentityCoreServiceComponent.getURLResolverService();
+        return getAbsoluteURL(urlContext, addProxyContextPath, addWebContextRoot, addTenantQueryParamInLegacyMode,
+                addTenantPathParamInLegacyMode);
+    }
+
+    public static String getAbsoluteURL(String urlContext, boolean addProxyContextPath, boolean addWebContextRoot,
```
malithie, line 460: Let's not have this as a public methos

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityUtil.java
```diff
@@ -451,12 +453,34 @@ public static String getServerURL(String urlContext, boolean addProxyContextPath
...
+                                        boolean addTenantQueryParamInLegacyMode, boolean addTenantPathParamInLegacyMode)
+            throws IdentityRuntimeException {
+
+        ServiceURLBuilder serviceURLBuilder = IdentityCoreServiceComponent.getServiceURLBuilder();
```
malithie, line 464: Can't we call, ServiceURLBuilder.create()

malithie: Intention of builder pattern is to write this code with less lines like,
ServiceURLBuilder.create().addPath().addParameter()

the respective API should validate for empty and null values

ChamathNS: The builder components doesn't take in any parameters. So, for example, ServiceURLBuilder.create().addPath().addTenantQueryParam() gets called, the ServiceURLBuilder registers this as addTenantQueryParam = true

Since we don't know the value of the booleans in getServerUrl method, we have to perform a check for the boolean parameters before adding it to the builder.

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
malithie: @ChamathNS not fixed still right ?

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+        return new DefaultServiceURLBuilder();
+    }
+
+    /**
+     * This method is called to add the URL path to the builder.
+     *
+     * @param paths Context paths. Can provide multiple context paths with a comma separated string.
+     * @return {@link ServiceURLBuilder}.
+     */
+    ServiceURLBuilder addPath(String paths);
```
malithie, line 40: Accept as varargs, so you don't need to tokenize

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
```diff
@@ -0,0 +1,98 @@
...
+     */
+    ServiceURLBuilder addParameter(String key, String value);
+
+    /**
+     * This method is called when the URL needs to be appended with a fragment.
+     *
+     * @param fragment Fragment.
+     * @return {@link ServiceURLBuilder}.
+     */
+    ServiceURLBuilder addFragment(String fragment);
```
malithie, line 88: I'm not exactly sure if we should have key value supported as well. An overloaded method may be.
@mefarazath we are referring to key value pairs in oauth for fragments right

malithie: From URL perspective there will be one fragment similar to query string. But we can have a method that accepts key, value pairs in a fragment, which might help building such fragments easily.

And above we can rename as setFragment which will denote that it will set the fragment component of the URL

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
ChamathNS: Should we add fragments to relative URL as well?

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+
+    /**
+     * This method is called when the URL fragment needs to be appended with parameters. Such parameters should be
+     * entered as <k,v> pairs. These parameters will get appended with an "&".
+     *
+     * @param key   Key.
+     * @param value Value.
+     * @return {@link ServiceURLBuilder}.
+     */
+    ServiceURLBuilder addFragmentAsParams(String key, String value);
```
malithie, line 68: We can add multiple params right
What if we say 'addFragmentParams'

malithie: or 'addFragmentParameter'

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilderFactory.java
```diff
@@ -0,0 +1,36 @@
...
+public class ServiceURLBuilderFactory {
+
+    /**
+     * This method is called to create a {@link ServiceURLBuilder} instance.
+     *
+     * @return {@link ServiceURLBuilder}.
+     */
+    public ServiceURLBuilder getServiceURLBuilderInstance() {
+
+        return ServiceURLBuilder.create();
```
malithie, line 33: Well the case is if somebody extends 'ServiceURLBuilder' still its create cannot be overloaded and the interface will return an implementation of DefaultURLServiceBuilder.

What if we consume the factory in ServiceURLBuilder.create() ?

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,188 @@
...
+     * entered as <k,v> pairs.
+     *
+     * @param key   Key.
+     * @param value Value.
+     * @return {@link ServiceURLBuilder}.
+     */
+    @Override
+    public ServiceURLBuilder addParameter(String key, String value) {
+
+        parameters = new HashMap<>();
```
malithie, line 75: Why create a new hashmap here. It should be pre-initialized.
This avoids invoking this method multiple times and adding several parameters

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,188 @@
...
+     * entered as <k,v> pairs. These parameters will get appended with an "&".
+     *
+     * @param key   Key.
+     * @param value Value.
+     * @return {@link ServiceURLBuilder}.
+     */
+    @Override
+    public ServiceURLBuilder addFragmentAsParams(String key, String value) {
+
+        fragmentParams = new HashMap<>();
```
malithie, line 104: Same above applies here

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,188 @@
...
+
+    private String resolveUrlPath(String[] urlPaths) {
+
+        StringBuilder urlPathBuilder = new StringBuilder();
+        if (ArrayUtils.isNotEmpty(urlPaths)) {
+            for (String path : urlPaths) {
+                urlPathBuilder.append(path).append("/");
+            }
+            if (urlPathBuilder.toString().endsWith("/")) {
+                urlPathBuilder.setLength(urlPathBuilder.length() - 1);
```
malithie, line 153: Is this the best way you can remove '/' from end. I think you can use charAt

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,188 @@
...
+
+    private String fetchHostName() throws URLBuilderException {
+
+        String hostName = ServerConfiguration.getInstance().getFirstProperty(IdentityCoreConstants.HOST_NAME);
+        try {
+            if (StringUtils.isBlank(hostName)) {
+                hostName = NetworkUtils.getLocalHostname();
+            }
+        } catch (SocketException e) {
+            throw new URLBuilderException("Error while trying to resolve the hostname from the system", e);
```
malithie, line 172: Better to log unresolvable host in message so that it makes the message more context aware

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilderFactory.java
```diff
@@ -0,0 +1,36 @@
...
+ * Service URL Builder Factory representation.
+ */
+public class ServiceURLBuilderFactory {
+
+    /**
+     * This method is called to create a {@link ServiceURLBuilder} instance.
+     *
+     * @return {@link ServiceURLBuilder}.
+     */
+    public ServiceURLBuilder getServiceURLBuilderInstance() {
```
malithie, line 31: Shall we rename as 'getServiceURLBuilder'

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilderFactory.java
malithie: This should return new DefaultServiceURLBuilder();

and in create() of 'ServiceURLBuilder' you should have,

if IdentityCoreServiceComponent.getServiceURLBuilderFactory() != null {

return IdentityCoreServiceComponent.getServiceURLBuilderFactory().getServiceURLBuilder();

else 

new DefaultServiceURLBuilder ()

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
```diff
@@ -0,0 +1,276 @@
...
+     * @return ArrayList of parameter keys.
+     */
+    public ArrayList<String> getParameterKeys() {
+
+        ArrayList<String> parameterArrayList = new ArrayList<>();
+
+        for (Map.Entry<String, String> entry : this.parameters.entrySet()) {
+            parameterArrayList.add(entry.getKey());
+        }
+        return parameterArrayList;
```
malithie, line 130: return an array

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
```diff
@@ -0,0 +1,276 @@
...
+     * @throws URLBuilderException
+     */
+    public String getFragment() throws URLBuilderException {
+
+        URI uri;
+        try {
+            uri = new URI(this.getAbsoluteUrl());
+        } catch (URISyntaxException e) {
+            throw new URLBuilderException("Error while trying to build the url", e);
+        }
```
malithie, line 146: Why build a URI. Why not return from fragment variable

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
ChamathNS: Looked into this and found that setting the length in the above manner is more efficient than using `stringbuilder.deleteCharAt(int index)`

ref: []()https://stackoverflow.com/a/10131077/6113775

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
ChamathNS: Went this way because the fragments are set in two ways. Either by setFragment(String fragment) where the fragment is set via a String, or by setFragmentParameters(String key, String value).. When the fragment is added via the second method, we don't store the fragment as a string before we encode it and append to the url. To resolve this, we can introduce a variable to store the fragment as a string before it is encoded and appended to the URL.

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
malithie: mmm but you are creating a string

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
malithie: setFragement denotes -> fragement is fully replaced
addFragementParameter -> denotes key value par is added as to the fragment and you can call it many times to add.

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
```diff
@@ -0,0 +1,84 @@
...
+public interface ServiceURLBuilder {
+
+    static ServiceURLBuilder create() {
+
+        if (IdentityCoreServiceComponent.getServiceURLBuilderFactory() != null) {
+            return IdentityCoreServiceComponent.getServiceURLBuilderFactory().getServiceURLBuilder();
+        } else {
+            return new DefaultServiceURLBuilder();
+        }
+
```
malithie, line 37: remove unnecessary line.
Also, hope we checked that factory gets resolved from OSGi service level

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilderFactory.java
```diff
@@ -0,0 +1,38 @@
...
+ * Service URL Builder Factory representation.
+ */
+public class ServiceURLBuilderFactory {
+
+    /**
+     * This method is called to create a {@link ServiceURLBuilder} instance.
+     *
+     * @return {@link ServiceURLBuilder}.
+     */
+    public ServiceURLBuilder getServiceURLBuilder() {
```
malithie, line 33: As this a factory wouldn't it be better if we say createServiceURLBuilder, else it does not denote at once that a new builder will be created

malithie: open for discussions @mefarazath

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
ChamathNS: Yes, created a dummy method and a test in IdentityUtil and checked this.

components/identity-core/org.wso2.carbon.identity.core/src/test/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilderTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+        String testPath = "/testPath";
+        ServiceURL serviceURL = null;
+        try {
+            serviceURL = ServiceURLBuilder.create().addPath(testPath).build();
+        } catch (URLBuilderException e) {
+            // Mock behaviour, hence ignored
+        }
+
+        assertNotNull(serviceURL.getUrlPath());
```
malithie, line 109: In each of these methods we should check if the URL is built as expected. Null check is not enough

malithie: basically verify for absolute url, relative url and the respective component like fragment, parameter etc.

ChamathNS: That is done in the ServiceURLTest class.

ref: https://github.com/wso2/carbon-identity-framework/blob/4d5669512d85c459c670d72dfe0692fa7dddeeda/components/identity-core/org.wso2.carbon.identity.core/src/test/java/org/wso2/carbon/identity/core/model/ServiceURLTest.java#L182

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
```diff
@@ -0,0 +1,252 @@
...
+    /**
+     * This method is called to get the host name of {@link ServiceURL}.
+     *
+     * @return String of the host name.
+     */
+    public String getHostName() {
+
+        if (hostName.endsWith("/")) {
+            hostName = hostName.substring(0, hostName.length() - 1);
+        }
```
malithie, line 81: Should this check be here ?
Ideally it hostname here should not set with '/'. It should be validated at the URL builder level

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
```diff
@@ -0,0 +1,252 @@
...
+    private String hostName;
+    private int port;
+    private String urlPath;
+    private Map<String, String> parameters;
+    private String fragment;
+    private String resolvedParamsString;
+    private String absoluteUrl;
+    private String relativeUrl;
+
+    public ServiceURL(String protocol, String hostName, int port, String urlPath, Map<String, String> parameters,
```
malithie, line 48: Having this public will be a problem. Looks like no option here though unless we move packages

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
ChamathNS: Yes, will move it to the DefaultServiceURLBuilder class.

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/model/ServiceURL.java
```diff
@@ -0,0 +1,252 @@
...
+
+        if (serverUrl.length() > 0 && serverUrl.charAt(serverUrl.length() - 1) == '/') {
+            serverUrl.setLength(serverUrl.length() - 1);
+        }
+        if (StringUtils.isNotBlank(resolvedParamsString)) {
+            try {
+                serverUrl.append(delimiter).append(URLEncoder.encode(resolvedParamsString,
+                        StandardCharsets.UTF_8.name()));
+            } catch (UnsupportedEncodingException e) {
+                throw new URLBuilderException("Error while trying to build the url", e);
```
malithie, line 188: Add more context info here

malithie: Shouldn't we return with an error code. @mefarazath what was the approach followed previously. Was it in API level we returned with error codes

components/identity-core/org.wso2.carbon.identity.core/src/test/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilderTest.java
malithie: Then why check null. Shouldn't the expectation is to verify the path added

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,213 @@
...
+        String tenantDomain = IdentityTenantUtil.getTenantDomainFromContext();
+        if (StringUtils.isBlank(tenantDomain)) {
+            tenantDomain = PrivilegedCarbonContext.getThreadLocalCarbonContext().getTenantDomain();
+        }
+        return tenantDomain;
+    }
+
+    private String buildFragment(String fragment, Map<String, String> fragmentParams) {
+
+        if (StringUtils.isNotBlank(fragment)) {
```
mefarazath, line 149: Can there be a situation where we will have both the fragment string and fragment params?

ChamathNS: The spec for URIs [1] says the semantics of the fragment are considered unknown and are effectively unconstrained. So, if we want to accommodate such a case, we can have it. But, we need to rethink the fragment structure then.. 

ref: []()https://tools.ietf.org/html/rfc3986#section-3.5

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURL.java
```diff
@@ -0,0 +1,94 @@
...
+     * @return The value of the parameter.
+     */
+    String getParameter(String key);
+
+    /**
+     * This method is called to get a list of the parameter names.
+     *
+     * @return ArrayList of parameter keys.
+     */
+    String[] getParameterKeys();
```
mefarazath, line 67: Can we make this a list or a set?

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURL.java
```diff
@@ -0,0 +1,94 @@
...
+     */
+    String getUrlPath();
+
+    /**
+     * This method is called to get the parameter value when the key is provided.
+     *
+     * @param key Key of the parameter.
+     * @return The value of the parameter.
+     */
+    String getParameter(String key);
```
mefarazath, line 60: Also would be useful to return all parameters as a map.

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURL.java
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * This method is called to get the relative url from the url context.
+     *
+     * @return The relative URL from the Service URL instance.
+     */
+    String getRelativeURL() throws URLBuilderException;
+
+
+
+
```
mefarazath, line 93: Looks like too many new lines at the end

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURL.java
```diff
@@ -0,0 +1,94 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.carbon.identity.core;
+
+/**
+ * Service URL interface.
+ */
+public interface ServiceURL {
```
mefarazath, line 24: How about an API to get the tenant domain?

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURL.java
ChamathNS: The list we return should be unmodifiable, yes?

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURL.java
```diff
@@ -0,0 +1,107 @@
...
+     * @return The parameters.
+     */
+    Map<String, String> getParameters();
+
+    /**
+     * This method is called to get a list of the parameter names.
+     *
+     * @return ArrayList of parameter keys.
+     */
+    List<String> getParameterKeys();
```
mefarazath, line 77: Since we return the param map, just wondering whether we would have an actual requirement to return the keys. WDYT?

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURL.java
```diff
@@ -0,0 +1,107 @@
...
+     * @return value of the port.
+     */
+    int getPort();
+
+    /**
+     * This method is called to get the Url path of Service URL.
+     *
+     * @return String of the url path.
+     */
+    String getUrlPath();
```
mefarazath, line 55: Would it make sense to name this as getPath()?
getPath() is a commonly used https://docs.oracle.com/javase/7/docs/api/java/net/URL.html#getPath()

components/identity-core/org.wso2.carbon.identity.core/src/test/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilderTest.java
```diff
@@ -0,0 +1,234 @@
...
+
+    }
+
+    @ObjectFactory
+    public IObjectFactory getObjectFactory() {
+
+        return new org.powermock.modules.testng.PowerMockObjectFactory();
+    }
+
+}
```
mefarazath, line 234: New line at EOF

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURL.java
malithie: yeh possibly not. let's remove this

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,426 @@
...
+            if (resolvedUrlContext.trim().charAt(0) != '/') {
+                resolvedUrlStringBuilder.append("/").append(resolvedUrlContext.trim());
+            } else {
+                resolvedUrlStringBuilder.append(resolvedUrlContext.trim());
+            }
+        }
+
+//        return new ServiceURL(protocol, hostName, port, resolvedUrlStringBuilder.toString(), parameters,
+//                resolvedFragment);
+
```
malithie, line 138: remove redundant code

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
malithie: Let's not have both. setFragment() denotes it overrides. Common practice followed

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,426 @@
...
+            }
+
+            /**
+             * Concatenate the protocol, host name, port, proxy context path, web context root and url context to
+             * return the absolute URL.
+             *
+             * @return The absolute URL from the Service URL instance.
+             */
+            @Override
+            public String getAbsoluteURL() throws URLBuilderException {
```
malithie, line 253: It's fine for the builder to throw URLBuilderException while building the URL. But ServiceURL POJO should not throw it when getting absolute URL or relative URL

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,426 @@
...
+                resolvedUrlStringBuilder.append("/").append(resolvedUrlContext.trim());
+            } else {
+                resolvedUrlStringBuilder.append(resolvedUrlContext.trim());
+            }
+        }
+
+//        return new ServiceURL(protocol, hostName, port, resolvedUrlStringBuilder.toString(), parameters,
+//                resolvedFragment);
+
+        return new ServiceURL() {
```
malithie, line 139: This is a long class to implement as an anonymous. Impacts readability

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURL.java
```diff
@@ -0,0 +1,107 @@
...
+     */
+    String getTenantDomain();
+
+    /**
+     * Concatenate the protocol, host name, port, proxy context path, web context root and url context to return the
+     * absolute URL.
+     *
+     * @return The absolute URL from the Service URL instance.
+     */
+    String getAbsoluteURL() throws URLBuilderException;
```
malithie, line 99: I don't think exceptions should be thrown from here

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURL.java
malithie: +1 to use getPath()

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURLBuilder.java
```diff
@@ -0,0 +1,82 @@
...
+    ServiceURLBuilder addFragmentParameter(String key, String value);
+
+    /**
+     * This method is used to return a ServiceURL with the protocol, hostname, port, proxy context path, a web context
+     * root and the tenant domain (appended if required).
+     *
+     * @return {@link ServiceURL}.
+     * @throws URLBuilderException If error occurred while constructing the URL.
+     */
+    ServiceURL build() throws URLBuilderException, ServiceURLException;
```
malithie, line 81: Let's stick to one exception type. No need to re-wrap and throw. These are not two specific types we should declare. Rather you have been wrapping the same thing.

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
```diff
@@ -0,0 +1,447 @@
...
+                } else {
+                    absoluteUrl.append(urlPath.trim());
+                }
+            }
+            String resolvedParamsString = getResolvedParamString(parameters);
+            try {
+                appendParamsToUri(absoluteUrl, resolvedParamsString, "?");
+                appendParamsToUri(absoluteUrl, fragment, "#");
+
+            } catch (URLBuilderException e) {
```
malithie, line 218: Why we need to wrap this and throw ?
Still this relates to the same thing right. Let's throw directly one type

PR #2869, opened 2020-04-11
Improve URL resolving logic in ConfigurationFacade

mefarazath:
Minor refactoring

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/ConfigurationFacade.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/ConfigurationFacade.java
@@ -44,8 +44,8 @@ import java.util.Map;
 import java.util.function.Supplier;
 
 import static org.wso2.carbon.identity.application.authentication.framework.util.FrameworkConstants.DefaultUrlContexts.AUTHENTICATION_ENDPOINT;
-import static org.wso2.carbon.identity.application.authentication.framework.util.FrameworkConstants.DefaultUrlContexts.AUTHENTICATION_ENDPOINT_MISSING_CLAIMS_PROMPT;
 import static org.wso2.carbon.identity.application.authentication.framework.util.FrameworkConstants.DefaultUrlContexts.AUTHENTICATION_ENDPOINT_DYNAMIC_PROMPT;
+import static org.wso2.carbon.identity.application.authentication.framework.util.FrameworkConstants.DefaultUrlContexts.AUTHENTICATION_ENDPOINT_MISSING_CLAIMS_PROMPT;
 import static org.wso2.carbon.identity.application.authentication.framework.util.FrameworkConstants.DefaultUrlContexts.AUTHENTICATION_ENDPOINT_RETRY;
 import static org.wso2.carbon.identity.application.authentication.framework.util.FrameworkConstants.DefaultUrlContexts.AUTHENTICATION_ENDPOINT_WAIT;
 import static org.wso2.carbon.identity.application.authentication.framework.util.FrameworkConstants.DefaultUrlContexts.IDENTIFIER_FIRST_CONFIRMATION;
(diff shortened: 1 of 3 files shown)


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/ConfigurationFacade.java
```diff
@@ -263,28 +266,60 @@ public int getMaxLoginAttemptCount() {
...
+                log.debug("URL from file based config: " + urlFromFileBasedConfig);
+            }
+            context = urlFromFileBasedConfig;
+        }
+
+        if (isUrlAbsolute(context)) {
+            if (log.isDebugEnabled()) {
+                log.debug("URL: " + context + " is absolute. Returning it as the resolved URL.");
+            }
+            return context;
```
malithie, line 288: There's a small concern here.
We should read from file only when it's not tenant url supported mode right.
If it's tenant url supported, context is always a path, that will be resolved to an absolute URL from the URL builder.

If tenant url not supported only we will have to worry about, the context read from the file is absolute or not.

mefarazath: Even in the tenant qualified mode, we might have a situation where a user has defined a custom path. For example: instead of `authenthenticationendpoint/login.do` someone could define `authenthenticationendpoint/custom_login.do`

So when the facade returns in the tenant qualified mode, we can simply return 
`\t\<tenant_name>\authenthenticationendpoint/custom_login.do`

malithie: I was actually thinking that should be resolved local to the tenant.
But, unlike other endpoints which are defined for the server, there's a case with related to authentication endpoint web app. That is, it allows to be externalized, which can mean server wide there can be an endpoint URL configured, where your above comment applies. Mode of the offering needs to be finalized then

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/ConfigurationFacade.java
```diff
@@ -263,28 +263,25 @@ public int getMaxLoginAttemptCount() {
...
-                // Use the value configured in the file.
-                url = authenticationEndpointURLInFile;
+            String urlFromFileBasedConfig = getValueFromFileBasedConfig.get();
+            if (StringUtils.isNotBlank(urlFromFileBasedConfig)) {
+                // If the file based URL is set, then we have to return the file based URL.
+                return urlFromFileBasedConfig;
             } else {
-                // Use the default context.
-                url = defaultContext;
+                return defaultContext;
```
malithie, line 274: Should we return the default context here.
Shouldn't that be a absolute public ?

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/ConfigurationFacade.java
```diff
@@ -263,28 +263,25 @@ public int getMaxLoginAttemptCount() {
...
     }
 
-    private String buildTenantQualifiedUrl(String context) {
+    private String buildAbsoluteUrl(String context) {
 
         try {
             return ServiceURLBuilder.create().addPath(context).build().getAbsolutePublicURL();
         } catch (URLBuilderException e) {
-            throw new IdentityRuntimeException("Error while building url for context: " + context);
+            throw new IdentityRuntimeException("Error while building tenant qualified url for context: " + context);
```
malithie, line 284: I would prefer if this exception can be thrown and handled at the public API layer

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/ConfigurationFacade.java
mefarazath: Though of building the absolute public at the consumer level to avoid behaviour change for code that consumes this API in the legacy mode

PR #2885, opened 2020-04-20
Obtain tenant domain for ServiceURLBuilder only from the thread local.

ChamathNS:

Update DefaultServiceURLBuilder to obtain the tenant domain only from the thread local.
improvement for #2851.

### DIFF
--- a/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
+++ b/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
@@ -23,7 +23,6 @@ import org.apache.commons.collections.MapUtils;
 import org.apache.commons.lang.ArrayUtils;
 import org.apache.commons.lang.StringUtils;
 import org.wso2.carbon.base.ServerConfiguration;
-import org.wso2.carbon.context.PrivilegedCarbonContext;
 import org.wso2.carbon.identity.core.ServiceURL;
 import org.wso2.carbon.identity.core.ServiceURLBuilder;
 import org.wso2.carbon.identity.core.URLBuilderException;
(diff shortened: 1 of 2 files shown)


components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
```diff
@@ -331,11 +330,7 @@ public ServiceURLBuilder addFragmentParameter(String key, String value) {
 
     private String resolveTenantDomain() {
 
-        String tenantDomain = IdentityTenantUtil.getTenantDomainFromContext();
-        if (StringUtils.isBlank(tenantDomain)) {
-            tenantDomain = PrivilegedCarbonContext.getThreadLocalCarbonContext().getTenantDomain();
-        }
-        return tenantDomain;
+        return IdentityTenantUtil.getTenantDomainFromContext();
```
mefarazath, line 333: In the non tenant qualified URL mode, shouldn't we get the tenantDomain from the context?

ChamathNS: In the non-tenant qualified mode, the ServiceURLBuilder does not obtain the tenant domain since it won't append that in the path.

malithie: I don't think we should do this change now.
We should have it for non tenant mode. Appending the tenant to URL should be decided based on the config not from the tenant in context right

ChamathNS: +1

malithie:
@ChamathNS should this be closed ?

PR #2908, opened 2020-04-29
Fix SAML metadata URLs in tenant qualified URL mode

mefarazath:

In the tenant qualified URL mode,

- Fixes SAML endpoint URLs build for SAML metadata being overridden by already saved values in the DB.
- Adds the SAML SSO URL as a destination URL by default in addition to already configured ones.

### DIFF
--- a/components/idp-mgt/org.wso2.carbon.idp.mgt/src/main/java/org/wso2/carbon/idp/mgt/IdentityProviderManager.java
+++ b/components/idp-mgt/org.wso2.carbon.idp.mgt/src/main/java/org/wso2/carbon/idp/mgt/IdentityProviderManager.java
@@ -79,6 +79,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.xml.stream.XMLStreamException;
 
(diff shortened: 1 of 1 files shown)


components/idp-mgt/org.wso2.carbon.idp.mgt/src/main/java/org/wso2/carbon/idp/mgt/IdentityProviderManager.java
```diff
@@ -702,6 +578,173 @@ public IdentityProvider getResidentIdP(String tenantDomain)
         return identityProvider;
     }
 
+    private String buildSAMLUrl(String urlFromConfigFile, String defaultContext, boolean appendTenantDomainInLegacyMode)
+            throws IdentityProviderManagementServerException {
+
+        String url = urlFromConfigFile;
+        if (StringUtils.isBlank(url)) {
+            url = IdentityUtil.getServerURL(defaultContext, true, true);
```
ashensw, line 586: Can't we use the ServiceURLBuilder here to get the URL?

mefarazath: Some of the SAML URLs don't need the tenant domain appended as a query param. For example the artifact resolve URL. 

If we use the ServiceURLBuilder all of them will be appended with the tenant domain. 

But i think we can improve and use two methods for the two cases.

PR #2941, opened 2020-05-18
Add resource context for lsp

Sathiyakugan:


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -1364,6 +1364,7 @@
             {% endfor %}
         </Resource>
         {% endfor %}
+        <Resource context="(.*)/lsp/(.*)" secured="true" http-method="all">
```
darshanasbg, line 1367: Need to define a permission level also.

PR #3002, opened 2020-07-06
[Fix]Local Roles are not deleted in JIT provisioning flow when the federated IDP roles are removed

inthirakumaaran:
fix wso2/product-is#8629

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/provisioning/impl/DefaultProvisioningHandler.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/provisioning/impl/DefaultProvisioningHandler.java
@@ -165,7 +165,7 @@ public class DefaultProvisioningHandler implements ProvisioningHandler {
                 }
             }
 
-            if (roles != null && !roles.isEmpty()) {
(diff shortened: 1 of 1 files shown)


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/provisioning/impl/DefaultProvisioningHandler.java
```diff
@@ -165,7 +165,7 @@ public void handle(List<String> roles, String subject, Map<String, String> attri
                 }
             }
 
-            if (roles != null && !roles.isEmpty()) {
+            if (roles != null) {
```
madurangasiriwardena, line 168: A little bit confused about what's happening here. Though we check `roles != null`, `roles` variable is not considered within the `if` block.

inthirakumaaran: This `roles` is used in the above lines to create the list of roles to add list, then that list is used inside this IF block to update the user's new role set. If the `roles` is null we don't update the user's roles.

Though the order seems a bit confusing `!roles.isEmpty()` check always become false if we have only user role and later remove that role.

IsuraD (approved):

PR #3003, opened 2020-07-06
Adding allowed-auth-handlers param for custom <Resource> under <ResourceAccessControl> in identity.xml.j2

Dinushi:

The identity.xml.j2 of IS-KM 5.10 doesn't support adding the ''allowed-auth-handlers" param for a custom Resource using the deployment.toml, which is a requirement of WSO2 Open Banking solution.

This fix will enable the support to add a list of auth-handlers for each custom \<Resource> under \<ResourceAccessControl> in identity.xml.

This can be configured from deployment.toml as follows.
```
[[resource.access_control]]
context = "<context_name>"
secure="true"
http_method="all"
permissions=["/permission/admin"]
allowed_auth_handlers = ["AuthHandler1", "AuthHandler2"]
```
Related Issue : https://github.com/wso2-enterprise/financial-open-banking/issues/5597
                         https://github.com/wso2-enterprise/financial-open-banking/issues/5598

can merge this pull request.]

### Follow up actions

environments.]

#### General

#### Functionality

#### Code

### DIFF
--- a/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
+++ b/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
@@ -1386,7 +1386,8 @@
 
     <ResourceAccessControl {% if resource_access_control.default_access_allow is sameas false %} default-access="deny" {% endif %} {% if resource_access_control.disable_scope_validation is sameas true %} disable-scope-validation="true" {% endif %}>
         {% for resource in resource.access_control %}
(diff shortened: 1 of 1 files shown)


janakamarasena:
Hi @Dinushi since majority case will be this template will not be used by many lets template this in a way to add the config only if present in deployment.toml file.
like this:  https://github.com/wso2/carbon-identity-framework/blob/a1d9795d268456d6bec2ae9ee581219f6531ef3a/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2#L1320-L1326

I think this will work
```
{% if resource.allowed_auth_handlers is defined %}  allowed-auth-handlers="{{resource.allowed_auth_handlers|join(', ')}}" {% endif %}
```

PR #3005, opened 2020-07-08
Confirmation code introspection API for lite user registration.

Pushpalanka:
### Confirmation code introspection API for lite user registration.


components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint.util/src/main/java/org/wso2/carbon/identity/mgt/endpoint/util/client/model/ExtendedUser.java
```diff
@@ -0,0 +1,103 @@
...
+ *  KIND, either express or implied.  See the License for the
+ *  specific language governing permissions and limitations
+ *  under the License.
+ *
+ */
+package org.wso2.carbon.identity.mgt.endpoint.util.client.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class ExtendedUser {
```
IsuraD, line 23: Why can't we use the existing user object

Pushpalanka: This backend supports an additional attribute 'isUsernameCaseSensitive' related to code, that is not present in the current User object. Currently, this attribute is not used, but the username validations against code can honor this attribute. This is exposing all the details provided from the backend in this API.

PR #3121, opened 2020-09-22
Fix wso2/product-is#9531: SaaS app authentication with tenant qualified URLs

omindu:
Fixes wso2/product-is#9531

With this fix, during SaaS app authentication, we consider the tenant domain in the authentication context by default. As a result if the login identifier of the user does not contain the tenant domain, the user will be authenticated against the application tenant domain.

The default behavior for SaaS app authentication will be changed with the fix. However the below configuration in deployment.toml will enable the previous authentication behavior where we considered login identifier without tenant domain are from super tenant.

```toml
[authentication]
enable_legacy_saas_mode = true
```


pulasthi7 (approved):

omindu:
PR builder passed https://jenkins-support-all.wso2.com/jenkins/job/DEV_PR_BUILDER/811/

PR #3202, opened 2020-10-29
Returning client error when trying to delete a read-only template

somindatommy:
resolves https://github.com/wso2/product-is/issues/9529

### Approach
Check whether the given `templateId` is related to a read-only template and return the client error.


IsuraD (approved):

IsuraD:
@somindatommy we need to document the newly added error code.

PR #3225, opened 2020-11-07
Add configs to disable refresh token expiry time extension on refresh token renewal

mevan-karu:

### DIFF
--- a/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json
+++ b/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json
@@ -50,7 +50,7 @@
   "oauth.token_validation.refresh_token_validity": "1d",
   "oauth.timestamp_skew": "0m",
   "oauth.token_renewal.renew_refresh_token": true,
-  "oauth.token_renewal.extend_refresh_token_expiry_time_on_renewal": false,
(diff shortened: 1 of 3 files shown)


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json
```diff
@@ -50,6 +50,7 @@
   "oauth.token_validation.refresh_token_validity": "1d",
   "oauth.timestamp_skew": "0m",
   "oauth.token_renewal.renew_refresh_token": true,
+  "oauth.token_renewal.extend_refresh_token_expiry_time_on_renewal": false,
```
emswbandara, line 53: shouldn't this be true?

mevan-karu: According to issue description, default behavior should be not extending the refresh token expiry time on token renewal. Since we are fixing this in the public shouldn't we disable this?

mevan-karu: Refresh token expiry time extension is enabled by default after having an offline discussion with @mefarazath, since according to the spec browser-based apps should only adhere to the spec.

IsuraD (approved):

PR #3226, opened 2020-11-07
validation for MappedAttributes

nisho06:
Resolves https://github.com/wso2/product-is/issues/9810

## Approach
While mapped attribute was empty, it is not validating it and allows the request and makes the mapped attribute as empty in the IS. Since mappedAttribute is mandatory, it cannot be empty. So I have validated it and thrown an error


somindatommy:
@nisho06 The framework services in the framework do not only serve the REST APIs. Therefore, running the test scenarios is not enough. The best option here is to run against all the integration tests.

somindatommy (approved):

PR #3243, opened 2020-11-11
Fix syntax error while searching roles in oracle db

piraveena:
Resolves https://github.com/wso2/product-is/issues/10439

Earlier the oracle sql query was like 

```
SELECT UM_ROLE_NAME FROM 
	(SELECT UM_ROLE_NAME  rownum AS rnum FROM 
		(SELECT UM_ROLE_NAME FROM UM_HYBRID_ROLE ORDER BY UM_ID DESC) 

WHERE UM_TENANT_ID=:UM_TENANT_ID;
AND UM_ROLE_NAME LIKE :UM_ROLE_NAME; 
AND rownum <= :END_INDEX;
        ) 
WHERE rnum > :ZERO_BASED_START_INDEX;
```

- Here the select query has two attributes`SELECT UM_ROLE_NAME rownum AS rnum`. They need to be splitted by `,` 

- And  `WHERE UM_TENANT_ID=:UM_TENANT_ID;` part is outside the `(SELECT UM_ROLE_NAME FROM UM_HYBRID_ROLE ORDER BY UM_ID DESC) ` query. So the `UM_TENANT_ID` column was not available outside the select query. So moved it inside the select query


sachiniWettasinghe (approved):

piraveena:
PR builder is success here https://github.com/wso2/carbon-identity-framework/pull/3243

PR #3280, opened 2020-12-07
Issue 5365: rename cookie for manage multi tenant session

FinalFBoss:

Fix issue https://github.com/wso2/product-is/issues/5365:
- custom cookie session name adding tenant name for mange multi tenant session
- inserted new methods to return new cookie name depending from tenant (storeAuthCookieTenant and getAuthCookieTenant)
uses this new methods where are call the olds

Modify from tag v5.17.5

### Follow up actions

Check multi session user of different tenant in same browser

### Checklist (for reviewing)

I left old code commented, remove them after merge.


pulasthi7:
Hi @FinalFBoss 

Sorry for the late reply.

We decided to do the same using a different approach, that is by qualifying the session cookie path with the tenant domain. That way we can prevent unnecessary cookies being sent from the browser which may exceed allowed header lengths (if there are several sessions in different domains).

The fix and the related issue can be found at https://github.com/wso2/product-is/issues/8975

Anyway, thanks for your contribution and would like to see you continue with more contributions in future.

PR #3283, opened 2020-12-09
Handle JIT provisioning when user operation workflows are defined

pamodaaw:
Handle the exceptions raised in
wso2/product-is#10651
wso2/product-is#10672


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/provisioning/impl/DefaultProvisioningHandler.java
```diff
@@ -69,6 +70,7 @@
 
     private static final Log log = LogFactory.getLog(DefaultProvisioningHandler.class);
     private static final String ALREADY_ASSOCIATED_MESSAGE = "UserAlreadyAssociated";
+    private static final String USER_WORKFLOW_ENGAGED_ERROR_CODE = "WFM-10001";
```
emswbandara, line 73: we can use ErrorMessages.ERROR_CODE_USER_WF_ALREADY_EXISTS.getCode instead of defining the constant here again

emswbandara (approved):

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/provisioning/impl/DefaultProvisioningHandler.java
pamodaaw: but if we do that, we need to add workflow dependencies to the framework which would result in transitive dependency issues

PR #3295, opened 2020-12-17
Fix getUserList issue when username list is changed via old listeners

inthirakumaaran:
Fix wso2/product-is#10616

### Approach
If usernames are changed in `returnUserNamesList` then reflect corresponding users in `returnUsersList`

### DIFF
--- a/components/identity-mgt/org.wso2.carbon.identity.mgt/src/main/java/org/wso2/carbon/identity/mgt/listener/IdentityUserNameResolverListener.java
+++ b/components/identity-mgt/org.wso2.carbon.identity.mgt/src/main/java/org/wso2/carbon/identity/mgt/listener/IdentityUserNameResolverListener.java
@@ -18,9 +18,12 @@
 
 package org.wso2.carbon.identity.mgt.listener;
 
+import org.apache.commons.collections.CollectionUtils;
 import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.wso2.carbon.identity.application.authentication.framework.exception.UserSessionException;
+import org.wso2.carbon.identity.application.authentication.framework.util.FrameworkUtils;
 import org.wso2.carbon.identity.core.AbstractIdentityUserOperationEventListener;
 import org.wso2.carbon.identity.core.util.IdentityCoreConstants;
 import org.wso2.carbon.identity.core.util.IdentityUtil;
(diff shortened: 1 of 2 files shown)


mevan-karu (approved):

components/identity-mgt/org.wso2.carbon.identity.mgt/src/main/java/org/wso2/carbon/identity/mgt/listener/IdentityUserNameResolverListener.java
```diff
@@ -797,6 +804,36 @@ public boolean doPreGetUserListWithID(String claimUri, String claimValue, List<U
...
+
+        // Reflect newly add users by listeners in returnUsersList
+        if (CollectionUtils.isNotEmpty(returnUserNamesList)) {
+            tempUserNamesList.addAll(returnUserNamesList);
+            tempUserNamesList.removeAll(returnInitialUserNamesList);
+            for (String username : tempUserNamesList) {
+                User newUser = new User();
+                newUser.setUsername(username);
+                try {
+                    newUser.setUserID(FrameworkUtils.resolveUserIdFromUsername(userStoreManager, username));
```
Thumimku, line 827: **resolveUserIdFromUsername**, This method can be returned an empty user-id right? Don't we need to check its emptiness?

inthirakumaaran: This won't send empty id. It will create a new id if user doesn't have one.  Even if it is empty we don't need to handle it here as our main intension is to get the username

Thumimku: ACK, that's why we don't throw the exception just logging.

Thumimku (approved):

components/identity-mgt/org.wso2.carbon.identity.mgt/src/main/java/org/wso2/carbon/identity/mgt/listener/IdentityUserNameResolverListener.java
inthirakumaaran: This won't send empty id. It will create a new id if user doesn't have one. Even if it is empty we don't need to handle it here as our main intension is to get the username

PR #3337, opened 2021-01-21
Add session data and token clean up scheduler scripts

DinikaSen:

Adds the session data cleanup and token cleanup scheduler scripts with the documentation.
Related to : https://github.com/wso2/product-is/issues/11021


DinikaSen:
Closing since this is now resolved with a different approach

nilasini:
@DinikaSen Can u please specify the approach which solve this? In that case are we good to resolve https://github.com/wso2/product-is/issues/11021

DinikaSen:
This requirement was for Asgardeo to execute the clean up scripts. It is now fulfilled using cron jobs and hence these scripts are not required anymore. Hence, the requirement mentioned in https://github.com/wso2/product-is/issues/11021 is also not required anymore, so we can close that as well.

PR #3371, opened 2021-02-08
Modify confirmation code cleanup script to clean up data older than 30 days

DinikaSen:

Modify confirmation code cleanup script to clean up data older than 30 days


rksk:
I believe we need to fix this in all the DB types.

Another suggestion is to add a `sleep` in the same way we have done in other cleanup procedures. So it will not lock the tables continuously and let normal flows go through.

IsuraD (approved):

PR #3415, opened 2021-02-27
Add multi attribute login service

Chathurangap688:
Resolves: https://github.com/wso2/product-is/issues/8899
### Purpose   
In current IS versions not provide any option to use values other than username as login identifier. Therefore users are required to use only username for authentication scenarios such as user-portal login.
And also users can recover there password only by providing the username. 

### Describe the improvement
There is an option in admin console under “Resident Identity Provider > Login Policies > Multi Attribute Login” to enable or disable the multi attribute login feature and admin can provide a list of claim URIs which are allowed for use as login identifier.
 Then users can use allowed attribute value in the tenant domain as the login identifier without using username. The login identifier is identified by comparing the given regex pattern of claims and then returns resolved username. 
When a user tries to recover the password, the user can use given identifiers without limiting only for username. 
And also there is another option to maintain the uniqueness of the multi attribute login allowed claim values in a tenant domain. To enable this feature admin need to add additional claim property as given below. 

Property Name | Property Value
-- | --
isUnique | ture

### Additional context
Needs to provide regex patterns for each claim which are allowed for multi-attribute-login by admins.
All regexes provided must have different patterns.

### How to reproduce
Resident Identity Provider > Login Policies > Multi Attribute Login
Enable the feature and add the multi attribute login allowed claim URIs.
Ex: http://wso2.org/claims/emailaddress,http://wso2.org/claims/telephone,http://wso2.org/claims/username
Note: If you enable this feature you need to add the username claim URI for login with username.


Chathurangap688:
Close to create new PR

PR #3420, opened 2021-03-04
Add method to get url that contains hostname and port only

piraveena:

As a part of https://github.com/wso2/product-is/issues/11380

### DIFF
--- a/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURL.java
+++ b/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURL.java
@@ -18,6 +18,8 @@
 
 package org.wso2.carbon.identity.core;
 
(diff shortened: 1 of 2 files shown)


components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/ServiceURL.java
```diff
@@ -122,4 +122,15 @@
...
+
+    /**
+     * Returns the proxy server url when the Identity Server is fronted with a proxy.
+     * Concatenate the protocol, host name, port and  return the public absolute URL.
+     *
+     * @return The public absolute URL without any url path.
+     */
+    default String getAbsolutePublicUrlWithoutPath() {
+
+        return null;
```
Pushpalanka, line 134: Would it be better to return an empty String, than returning null?

Pushpalanka (approved):

PR #3421, opened 2021-03-04
Fix SCIM2 roles claim mapping

tharindu-bandara:
Fixes https://github.com/wso2/product-is/issues/11347.

Set default SCIM2 roles claim mapped to the wso2.roles claim.


tharindu-bandara:
Review comments:

- [ ] Use migration client to do the mapping

IsuraD (approved):

PR #3471, opened 2021-04-07
Revert breaking changes with group role separation

tharindu-bandara:
When group-role separation is disabled, still these changes affect the previous server-status thus removing


tharindu-bandara:
This is fixed in a separate effort

PR #3478, opened 2021-04-20
Add config to keep non-jit provisioned roles of a user during jit update

nipunthilakshan:
**Resolves** : wso2/product-is#7203

**Approach** : Considered idp role mappings and allow any local role which is not mapped to an external role not to be touched during JIT provisioning.

To enable this following configuration need to be added in deployment.toml file.

```
[idp_role_management]
return_only_mapped_local_roles = true
return_manually_added_local_roles = true
```

### DIFF
--- a/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
+++ b/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
@@ -1240,6 +1240,9 @@
     {% if idp_role_management.return_only_mapped_local_roles is defined %}
         <FederatedRoleManagement>
             <ReturnOnlyMappedLocalRoles>{{idp_role_management.return_only_mapped_local_roles}}</ReturnOnlyMappedLocalRoles>
+            {% if idp_role_management.return_manually_added_local_roles is defined %}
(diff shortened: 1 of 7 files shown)


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -1240,6 +1240,9 @@
     {% if idp_role_management.return_only_mapped_local_roles is defined %}
         <FederatedRoleManagement>
             <ReturnOnlyMappedLocalRoles>{{idp_role_management.return_only_mapped_local_roles}}</ReturnOnlyMappedLocalRoles>
+            {% if idp_role_management.return_manually_added_local_roles is defined %}
+            <ReturnManuallyAddedLocalRoles>{{idp_role_management.return_manually_added_local_roles}}</ReturnManuallyAddedLocalRoles>
+            {% endif %}
```
Kanapriya, line 1245: Don't need to be added to the default.json file? what would be the default value for this config?

nipunthilakshan: Addressed.

Kanapriya (approved):

PR #3485, opened 2021-04-21
Fix incorrect audit log for group creation

emswbandara:
Fixes https://github.com/wso2-enterprise/asgardeo-product/issues/3156

### DIFF
--- a/components/user-mgt/org.wso2.carbon.user.mgt/src/main/java/org/wso2/carbon/user/mgt/listeners/UserManagementAuditLogger.java
+++ b/components/user-mgt/org.wso2.carbon.user.mgt/src/main/java/org/wso2/carbon/user/mgt/listeners/UserManagementAuditLogger.java
@@ -27,6 +27,7 @@ import org.slf4j.MDC;
 import org.wso2.carbon.CarbonConstants;
 import org.wso2.carbon.identity.core.AbstractIdentityUserOperationEventListener;
 import org.wso2.carbon.identity.core.util.IdentityCoreConstants;
+import org.wso2.carbon.identity.core.util.IdentityUtil;
 import org.wso2.carbon.user.api.Permission;
 import org.wso2.carbon.user.core.UserStoreManager;
 import org.wso2.carbon.user.mgt.listeners.utils.ListenerUtils;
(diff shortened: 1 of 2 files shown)


ashensw (approved):

components/user-mgt/org.wso2.carbon.user.mgt/src/main/java/org/wso2/carbon/user/mgt/listeners/UserManagementAuditLogger.java
```diff
@@ -209,8 +210,13 @@ public boolean doPostAddRole(String roleName, String[] userList, Permission[] pe
                 JSONArray permissionsArray = new JSONArray(permissions);
                 dataObject.put(ListenerUtils.PERMISSIONS_FIELD, permissionsArray);
             }
-            audit.warn(createAuditMessage(ListenerUtils.ADD_ROLE_ACTION,
-                    ListenerUtils.getEntityWithUserStoreDomain(roleName, userStoreManager), dataObject, SUCCESS));
+            if (IdentityUtil.isGroupsVsRolesSeparationImprovementsEnabled()) {
```
tharindu-bandara, line 213: 1. `IdentityUtil.isGroupsVsRolesSeparationEnabled` = Product is able to separate groups from roles. Previously known Roles without prepended domain considered as groups, and if not considered as roles.
2. `IdentityUtil.isGroupsVsRolesSeparationImprovementsEnabled ` = Sepraration is reflected in Oauth, SCIM flows. And claims level.

I think for this we should use 1. But we need to check the prefixed domain alongside with this fix to differentiate roles from groups

tharindu-bandara: @ashensw wdyt?

ashensw: This listener is not applicable for hybrid roles (New **Roles**) AFAIR. We have separate listeners for hybrid roles. Better to verify that. If that's the case we can assume all the roles come here as **Groups** as for our new convention. In that case no need to check the prefixed domain IMO?

Regarding the config ideally, we will enable both in the product. But +1 to move to `IdentityUtil.isGroupsVsRolesSeparationEnabled` considering the SCIM level separation is done based on the config 1.

tharindu-bandara: > This listener is not applicable for hybrid roles (New **Roles**) AFAIR. We have separate listeners for hybrid roles. Better to verify that. If that's the case we can assume all the roles come here as **Groups** as for our new convention. In that case no need to check the prefixed domain IMO?
> 
+1.

emswbandara: @tharindu-bandara Couldn't find isGroupsVsRolesSeparationEnabled() in master branch
https://github.com/wso2/carbon-identity-framework/blob/master/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityUtil.java

Is this in a different class?

tharindu-bandara (approved):

components/user-mgt/org.wso2.carbon.user.mgt/src/main/java/org/wso2/carbon/user/mgt/listeners/UserManagementAuditLogger.java
emswbandara: As per offline discussions decided to keep this config as it is since we don't have the mentioned config.

PR #3536, opened 2021-05-21
Add support to configure portal hostnames and paths when tenant qualified urls enabled

mevan-karu:
This PR adds support to configure hostname and path for authentication-portal and recovery-portal when tenant qualified URLs are enabled. Following deployment.toml configurations can be used to change the hostnames and paths of authentication-portal and recovery-portal.

```
[authenticationendpoint]
hostname ="local.accounts.wso2.com"
path ="/authenticationendpoint"

[recoveryendpoint]
hostname ="local.accounts.wso2.com"
path ="/recoveryendpoint"
```

### DIFF
--- a/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
+++ b/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
@@ -83,6 +83,10 @@ public class DefaultServiceURLBuilder implements ServiceURLBuilder {
         String protocol = fetchProtocol();
         String proxyHostName = fetchProxyHostName();
         String internalHostName = fetchInternalHostName();
+        String authenticationEndpointHostName = fetchAuthenticationEndpointHostName();
+        String authenticationEndpointPath = fetchAuthenticationEndpointPath();
+        String recoveryEndpointHostName = fetchRecoveryEndpointHostName();
+        String recoveryEndpointPath = fetchRecoveryEndpointPath();
         int proxyPort = fetchPort();
         int transportPort = fetchTransportPort();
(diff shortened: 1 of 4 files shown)


components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/DefaultServiceURLBuilder.java
```diff
@@ -83,13 +83,29 @@ public ServiceURL build() throws URLBuilderException {
...
+                    urlPath.contains(authenticationEndpointPath)) {
+                return new ServiceURLImpl(protocol, authenticationEndpointHostName, internalHostName, proxyPort,
+                        transportPort, tenantDomain, proxyContextPath, urlPath, parameters, resolvedFragment);
+            }
+            if (recoveryEndpointHostName != null && recoveryEndpointPath != null &&
+                    urlPath.contains(recoveryEndpointPath)) {
+                return new ServiceURLImpl(protocol, recoveryEndpointHostName, internalHostName, proxyPort,
+                        transportPort, tenantDomain, proxyContextPath, urlPath, parameters, resolvedFragment);
+            }
+        }
```
ChamathNS, line 108: IMO this should be handled at the places where the recovery endpoint path and the authentication endpoint path are there in the URLs. Instead of handling it at the ServiceURLBuilder, can't we include this logic at those specific places? WDYT @mevan-karu @mefarazath ?

mevan-karu: @ChamathNS but according to the current implementation of the ServiceURLBuilder, we can't configure the hostname from outside. Hostname resolving is done internally by the ServiceURLBuilder.

IsuraD (approved):

PR #3594, opened 2021-07-01
Fix behaviour of return_only_mapped_local_roles property

vihanga-liyanage:
When the property idp_role_management.return_only_mapped_local_roles is added, the expectation would be to only send the IDP mapped roles to SP at all times. But at the moment, if a role mapping is not added or at least one matching mapping is not found, we're sending all IDP roles to SP. This PR fixes that issue.

Refers https://github.com/wso2/product-is/issues/11941


sachiniWettasinghe (approved):

vihanga-liyanage:
Need to discuss and finalize the fix. Hence closing the PR.

PR #3609, opened 2021-07-14
Configure BasicClientAuthenticationHandler for introspection endpoint.

Thisara-Welmilla:
Configure BasicClientAuthenticationHandler for introspection endpoint to authenticate request using client credentials.

### DIFF
--- a/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
+++ b/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
@@ -1833,7 +1833,7 @@
         <Resource context="(.*)/oauth2/introspect(.*)"
                 secured="{{resource_access_control.introspect.secured}}"
                 http-method="all"
-                allowed-auth-handlers="BasicClientAuthenticationHandler">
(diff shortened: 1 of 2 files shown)


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -1832,7 +1832,8 @@
         </Resource>
         <Resource context="(.*)/oauth2/introspect(.*)"
                 secured="{{resource_access_control.introspect.secured}}"
-                http-method="all">
+                http-method="all"
+                allowed-auth-handlers="BasicClientAuthenticationHandler">
```
emswbandara, line 1836: value should be configurable from deployment.toml. We need to read the default values from default.json

emswbandara: This should be similar to how the value of 'secured' element is read.
secured="{{resource_access_control.introspect.secured}}

Thisara-Welmilla:
Decide to not to change the existing introspection endpoint authentication handlers for is-product

PR #3240, opened 2020-11-11
OIDC Federated Authenticator configs set to always be overwritten in tenant qualified url mode

ashendes:

PR #3710, opened 2021-09-03
set IdPEntityId to be tenant qualified

somindatommy:
In the current implementation, even though `tenant-qualified URLs` is enabled, IdPEntityId did not reflect that. This PR changes that by overwriting with tenant.

Related to the improvements in https://github.com/wso2/carbon-identity-framework/pull/3240

### DIFF
--- a/components/idp-mgt/org.wso2.carbon.idp.mgt/src/main/java/org/wso2/carbon/idp/mgt/IdentityProviderManager.java
+++ b/components/idp-mgt/org.wso2.carbon.idp.mgt/src/main/java/org/wso2/carbon/idp/mgt/IdentityProviderManager.java
@@ -360,7 +360,7 @@ public class IdentityProviderManager implements IdpManager {
         }
         propertiesList = new ArrayList<Property>();
 
-        Property idPEntityIdProp = resolveFedAuthnProperty(getOIDCResidentIdPEntityId(), oidcFedAuthn,
+        Property idPEntityIdProp = resolveFedAuthnProperty(oauth2TokenEPUrl, oidcFedAuthn,
                 OPENID_IDP_ENTITY_ID);
         propertiesList.add(idPEntityIdProp);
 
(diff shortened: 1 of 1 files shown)


janakamarasena (approved):

mevan-karu (approved):

components/idp-mgt/org.wso2.carbon.idp.mgt/src/main/java/org/wso2/carbon/idp/mgt/IdentityProviderManager.java
```diff
@@ -360,7 +360,7 @@ public IdentityProvider getResidentIdP(String tenantDomain)
         }
         propertiesList = new ArrayList<Property>();
 
-        Property idPEntityIdProp = resolveFedAuthnProperty(getOIDCResidentIdPEntityId(), oidcFedAuthn,
+        Property idPEntityIdProp = resolveFedAuthnProperty(oauth2TokenEPUrl, oidcFedAuthn,
```
ashendes, line 363: Ideally, the fix should be done in the private `getOIDCResidentIdPEntityId` method. The URL should be resolved by calling `resolveAbsoluteURL` within the method.

ashendes: Please check how it would affect the other usages of the method as well.

somindatommy: If we fix it in a private method, it will be persisted in the DB. So when the `tenant-qualified URLs` is disabled, still the value will have the tenanted path. Is this expected?

ashendes: If thats the case, we can introduce another variable similar to the other property variables and resolve the URL for the returned value of this method. 
This would be better because there could be instances where the `idpEntityId` is not always equivalent to the OAuth2 token endpoint value. In the default product, the Entity ID is a customizable value.

ashendes (approved):

PR #3735, opened 2021-09-17
Improve function library validations

janakamarasena:
Improve function library validations

When adding the loop validation code we checked the best place to have the code. This code would be used in application management when adding/updating an SP and also in functional script management when adding/updating a script. The options considered for the placement of the common code was. 

- Extract nashorn related code to a new component - But with the new improvements the function library will not have any nashorn related code
- Add to Identity Base - Identity base is a common dependency for both components and contains a validation utility class. But the validation utilities currently available there serves a broader scope and adding the loop validation here will not have any usage for other components.
- Adding application management common as a dependency to functional scripts component - Since logically functional scripts will only be used in applications, it seemed correct to go with this option and have the validation code in the application management common component.

### DIFF
--- a/components/application-mgt/org.wso2.carbon.identity.application.common/src/main/java/org/wso2/carbon/identity/application/common/util/IdentityApplicationManagementUtil.java
+++ b/components/application-mgt/org.wso2.carbon.identity.application.common/src/main/java/org/wso2/carbon/identity/application/common/util/IdentityApplicationManagementUtil.java
@@ -68,6 +68,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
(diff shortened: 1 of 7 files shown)


components/application-mgt/org.wso2.carbon.identity.application.common/src/main/java/org/wso2/carbon/identity/application/common/util/IdentityApplicationManagementUtil.java
```diff
@@ -255,6 +263,8 @@ private IdentityApplicationManagementUtil() {
         samlAuthnContextComparisonLevelList.add(IdentityApplicationConstants.SAML2.AuthnContextComparison.MAXIMUM);
         samlAuthnContextComparisonLevelList.add(IdentityApplicationConstants.SAML2.AuthnContextComparison.BETTER);
         samlAuthnContextComparisonLevels = Collections.unmodifiableList(samlAuthnContextComparisonLevelList);
+
+        JS_LOOP_PATTERN = Pattern.compile("\\b(for|while|forEach)\\b");
```
madurangasiriwardena, line 267: Let's move this outside the static block.

janakamarasena: done

madurangasiriwardena (approved):

PR #3749, opened 2021-09-22
Fix table column length exceeding when adding claim properties

somindatommy:
Validate the length of the claim properties added to the `IDN_CLAIM_PROPERTY` table so that they does not exceed max of 255 chars.


JKAUSHALYA (approved):

somindatommy:
Merging this since the related test failures are not due to this

PR #3752, opened 2021-09-23
Add schemas configurations

emswbandara:
Relates to https://github.com/wso2/identity-api-server/pull/316


IsuraD (approved):

AnuradhaSK:
If the following config is applied to the `deployment.toml` file, 

- Enterprise user schema SCIM claims mapped to identity claims will be removed
- Move "urn:ietf:params:scim:schemas:core:2.0:User:phoneNumbers.pager",  
"urn:ietf:params:scim:schemas:core:2.0:User:phoneNumbers.fax" from core schema to user schema..
These two were defined in core schema because there was a mistake in claim-config.xml. If we change the schemas.xml.j2 in the correct way need a migration.

```
[schemas]
profile = "default"
```

PR #3783, opened 2021-10-05
Update the exsisiting claim check to ignore case.

Wathsara:

There was a possible unique key violation when we are trying to add claim that already in the database with a new case lead.


emswbandara (approved):

IsuraD (approved):

Wathsara:
To verify the build.

image

PR #3806, opened 2021-10-26
Improve jit provisioning feature

Thisara-Welmilla:
This fix introduce following changes:
- Prevent username claim update for jit provisioned user in federated IDP claims and local claim syncing process.
- Maintain the consistancancy for the format of username claim variable in DefaultProvisioningHandler.class and JITProvisioningPostAuthenticationHandler.class.


Thisara-Welmilla:
Decided to not apply these changes

PR #3834, opened 2021-11-24
Fix the issue with JWT signature

nilasini:

Resolves https://github.com/wso2-enterprise/asgardeo-product/issues/8017
After checking the complexity currently we are not introducing any libraries to generate JWT instead we are fixing the issues in the existing JWT generation mechanism.

With this PR expiry time will be calculated correctly, JWT will contain the base64 url encoded values without padding as mentioned in the spec [1]

> Base64url Encoding
>       Base64 encoding using the URL- and filename-safe character set
>       defined in Section 5 of RFC 4648 [RFC4648], with all trailing '='
>       characters omitted (as permitted by Section 3.2) and without the
>       inclusion of any line breaks, whitespace, or other additional
>       characters.  Note that the base64url encoding of the empty octet
>       sequence is the empty string.  (See Appendix C for notes on
>       implementing base64url encoding without padding.)

further in the current implementation we are specifying the algorithm as HS256 (HMAC-SHA256) [2] but actually we are using HmacSHA1, with this PR I have corrected that as well.

[1] https://datatracker.ietf.org/doc/html/rfc7515#section-2
[2] https://github.com/wso2/carbon-identity-framework/blob/d6ea69f541e5eb0257919b72e639c7c523fc0b43/components/application-mgt/org.wso2.carbon.identity.application.common/src/main/java/org/wso2/carbon/identity/application/common/util/IdentityApplicationManagementUtil.java#L748

### DIFF
--- a/components/application-mgt/org.wso2.carbon.identity.application.common/src/main/java/org/wso2/carbon/identity/application/common/util/IdentityApplicationManagementUtil.java
+++ b/components/application-mgt/org.wso2.carbon.identity.application.common/src/main/java/org/wso2/carbon/identity/application/common/util/IdentityApplicationManagementUtil.java
@@ -18,8 +18,6 @@
 package org.wso2.carbon.identity.application.common.util;
 
 import org.apache.axiom.om.OMElement;
-import org.apache.axiom.om.util.Base64;
-import org.apache.axiom.util.base64.Base64Utils;
 import org.apache.commons.lang.ArrayUtils;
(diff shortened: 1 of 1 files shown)


janakamarasena (approved):

nilasini:
Build fails with only one test case "SAMLFederationWithFileBasedSPAndIDPTestCase"[1], same test case has been passed locally hence merging this PR.

[1] https://github.com/wso2/product-is/actions/runs/1515652470

PR #3854, opened 2021-12-15
Change self registration callback URL

kayathiri4:
Change self registration callback url to `.*`


kayathiri4:
We should not handle the callback url change here. Hence, closing this PR.

PR #3907, opened 2022-02-08
Fix NullPointerException while updating SAML IdP

kayathiri4:
 "Password Protected Transport" is added to Authentication Context Class by default even if the Authentication Context is set to No. This breaks the older IDP setups. This PR fixes the above mentioned issue.
 
Related issue: https://github.com/wso2/product-is/issues/13038


piraveena:
@kayathiri4 can you trigger PR builder and merge it once PR builder is successful?

rajithacharith (approved):

PR #3917, opened 2022-02-11
Make authenticationOptions parameter case insensitive

ChanikaRuchini:
Resolves
-  https://github.com/wso2-enterprise/asgardeo-product/issues/9156

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/model/graph/JsGraphBuilder.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/model/graph/JsGraphBuilder.java
@@ -407,7 +407,7 @@ public class JsGraphBuilder {
             if (StringUtils.isNotBlank(idp)) {
                 filteredOptions.putIfAbsent(idp, new HashSet<>());
                 if (StringUtils.isNotBlank(authenticator)) {
-                    filteredOptions.get(idp).add(authenticator);
+                    filteredOptions.get(idp).add(authenticator.toLowerCase());
                 }
             }
         });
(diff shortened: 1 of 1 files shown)


vihanga-liyanage (approved):

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/model/graph/JsGraphBuilder.java
```diff
@@ -442,7 +442,7 @@ protected void filterOptions(Map<String, Map<String, String>> authenticationOpti
                             .getInstance().getLocalAuthenticators();
                         for (LocalAuthenticatorConfig localAuthenticatorConfig : localAuthenticators) {
                             if (authenticatorConfig.getName().equals(localAuthenticatorConfig.getName()) &&
-                                authenticators.contains(localAuthenticatorConfig.getDisplayName())) {
+                                authenticators.contains(localAuthenticatorConfig.getDisplayName().toLowerCase())) {
```
mefarazath, line 445: All authenticator classes usually have two names, an internal name and a display name (a.k.a display name).
For example for TOTP[1],

getName() --> "totp"
getFriendlyName() --> "TOTP"

Looking at the code it could be that we are filtering based on the friendly name and not the internal name. So I am skeptical whether this fix has any value if we are intending to use the friendly name in the adaptive scripts.


[1] https://github.com/wso2-extensions/identity-outbound-auth-totp/blob/18d9f4608f83efdd8e3ba7f6ecb2733d8fbdd14b/component/authenticator/src/main/java/org/wso2/carbon/identity/application/authenticator/totp/TOTPAuthenticator.java#L469-L489

ChanikaRuchini: Authenticator is given by display name. So the filtering is done based on that. As some authenticator display names are in uppercase and some are in lower case the suggested solution for this issue is to make this authenticationOptions parameter case insensitive.

PR #3926, opened 2022-02-19
Validate for exiting local claim mapping within same dialect when adding external claims

wenodpathirana:
- Fixes https://github.com/wso2/product-is/issues/13054 by introducing additional validation to verify the existing local claim mappings within the same dialect.
- New claim management error code `CMT-60004` is introduced for the validation errors.


rksk:
In my opinion, a DB level constraint is also required since two concurrent requests might bypass the Java level validation.

wenodpathirana:
> 

@rksk uniqueness cannot be achieved within the mapping table, since we allow same local claim mapping across multiple dialects.

wenodpathirana:
PR Builder is failing due to the integration test failures, where test class is adding an external claim with a local claim mapping which is already mapped to an existing external claim in the same dialect. 
Since this PR added the backend validation to prevent above scenario, adding external claim will fail. 

Integrations tests are fixed in https://github.com/wso2/product-is/pull/13157.

piraveena (approved):

PR #3986, opened 2022-04-06
Add SSO URL as the destination URL during tenant creation

piraveena:

Resolves https://github.com/wso2/product-is/issues/13037

When a new tenant is created, the destination URL is not added to the DB. 
- But in the frontend, we show the SSO URL as the destination URL when the destination URL is empty. https://github.com/wso2/carbon-identity-framework/blob/v5.20.363/components/idp-mgt/org.wso2.carbon.idp.mgt.ui/src/main/resources/web/idpmgt/idp-mgt-edit-local.jsp#L136

Screenshot 2022-04-06 at 14 05 03
Even though the UI shows above URL as the destination URL, it is not added to the DB. 

- In the SAML component https://github.com/wso2-extensions/identity-inbound-auth-saml/blob/v5.9.2/components/org.wso2.carbon.identity.sso.saml/src/main/java/org/wso2/carbon/identity/sso/saml/util/SAMLSSOUtil.java#L661, when the destination URL is empty, it takes the default IdP `/samlsso` endpoint as the destination URL.
- This makes inconsistency. UI shows `/samlsso?tenantDomain=def.com` as the destination URL. But it will fail if we try with that destination URL

In this PR, when a tenant is created the SSO URL will be added as the destination URL in the DB. Now UI and the DB will be consistent.

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/idp-mgt/org.wso2.carbon.idp.mgt/src/main/java/org/wso2/carbon/idp/mgt/IdentityProviderManager.java
+++ b/components/idp-mgt/org.wso2.carbon.idp.mgt/src/main/java/org/wso2/carbon/idp/mgt/IdentityProviderManager.java
@@ -644,10 +644,8 @@ public class IdentityProviderManager implements IdpManager {
         }
         propertiesList.add(idPEntityIdProperty);
 
-        if (IdentityTenantUtil.isTenantQualifiedUrlsEnabled()) {
-            // Add SSO URL as a destination URL if not already available.
-            addSSOUrlAsDestinationUrl(samlFederatedAuthConfig, samlSSOUrl, propertiesList);
-        }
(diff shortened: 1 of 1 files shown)


nilasini (approved):

mefarazath:
@piraveena will this work as expected in both tenant qualified URL enabled/disabled scenarios?

piraveena:
@mefarazath The issue was reported for tenant qualified URLs disabled case. In that case, we are not storing the destination URLs in the DB during tenant creation flow. 

But for tenant qualified case, we already store it. Now I removed the if condition for tenant qualified check, this will make sure the flow will work for tenant qualified disabled case too.

PR #3996, opened 2022-04-18
Fix consent filtering issue for openid scope

ImalshaG:
Fixes: https://github.com/wso2-enterprise/asgardeo-product/issues/5011

This PR adds a new flag to enable consent filtering based on the scope to support backward compatibility, and fixes the consent filtering issue when the scope was set to openid. This flag is set to true by default and will not send mandatory claims if they are not included in the requested scope. If a user needs to get mandatory claims regardless of the scope, this flag should be set to false.

### DIFF
--- a/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json
+++ b/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json
@@ -1,6 +1,7 @@
 {
   "authentication.consent.prompt": true,
   "authentication.consent.subject.prompt": true,
+  "authentication.consent.scope.filtering": true,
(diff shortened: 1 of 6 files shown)


senthalan (approved):

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json
```diff
@@ -2,6 +2,7 @@
   "authentication.consent.prompt": true,
   "authentication.consent.subject.prompt": true,
   "authentication.map_federated_users_to_local": false,
+  "authentication.enable_scope_based_claim_filtering": true,
```
pulasthi7, line 5: is this also added to the migration guide? since you are changing the default behaviour, this should be mentioned properly

vihanga-liyanage (approved):

PR #4016, opened 2022-04-27
Give support to Encode/Decode adaptive authentication script

sahandilshan:
### Purpose
* Give support to encode & decode adaptive authentication using base64 encoding.
* Fix https://github.com/wso2/product-is/issues/13545


somindatommy (approved):

senthalan:
As the changes here are only to the JSPs and do not have any behaviour change to the system, we don't need to have a configuration.

senthalan (approved):

PR #4052, opened 2022-06-01
Fix federated IDP logout issue after force authentication of another service provider

ThaminduDilshan:
During the logout flow, DefaultLogoutRequestHandler.handle() method tries to find out whether there's a previous authenticated session in the context and if found only, it calls the authenticator.process() method. When processing the authenticator, if it's a logout request, the authenticator will call the initiateLogoutRequest method of the SAMLSSOAuthenticator which will then send logout requests for external IDPs.

Since force authentication skips the existing session and creates a new one, inside the authentication context, new session will only be associated with the force authenticated app (OIDC app). Hence when it tries to retrieve previous authenticated sequence using the application which sends the logout request (SAML app), it won't be able to find the IDP (which is associated with the previous session). Therefore the authenticator process method will not be called for the federated idp.

In the SAML logout flow it can be observed that the federated IDP is existing as a previous authenticated IDP in the context. So as a fix, from the DefaultLogoutRequestHandler, if we cannot find a previous authenticated session, but previous authenticated IDPs are existing in the authentication context, we can try to call authenticator.process() function for each IDP. This will initiate logout flow for the SAML federated IDP.

#### Related Issues
- https://github.com/wso2/product-is/issues/13696

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/context/AuthenticationContext.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/context/AuthenticationContext.java
@@ -756,6 +756,21 @@ public class AuthenticationContext extends MessageContext implements Serializabl
     }
 
     /**
+     * Check whether the authenticator is already logged out.
+     *
+     * @param idpName Identity provider name.
+     * @param authenticatorName Authenticator name.
+     * @return true if the authenticator already logged out. False otherwise.
+     */
+    public boolean isLoggedOutAuthenticator(String idpName, String authenticatorName) {
+
+        if (loggedOutAuthenticators.containsKey(idpName)) {
+            return loggedOutAuthenticators.get(idpName).contains(authenticatorName);
+        }
+        return false;
+    }
+
+    /**
      * Clears all currently logged out authenticators from the context.
      */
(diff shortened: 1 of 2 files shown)


senthalan (approved):

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/context/AuthenticationContext.java
```diff
@@ -725,4 +727,54 @@ public void setLoginTenantDomain(String loginTenantDomain) {
 
         this.loginTenantDomain = loginTenantDomain;
     }
+
+    /**
+     * Get currently logged out authenticators with IDPs.
+     * @return logged out authenticators.
+     */
+    public Map<String, List<String>> getLoggedOutAuthenticators() {
```
senthalan, line 735: we can remove this method now

senthalan (approved):

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/context/AuthenticationContext.java
ThaminduDilshan: Fixed

PR #4136, opened 2022-07-20
Improve pom by removing version in sub component.

Thumimku:
$sub


piraveena (approved):
do we have this artifact in orbit? we can delete it from orbit since we decided to not to include in our distribution

Thumimku:
> do we have this artifact in orbit? Let's delete it from orbit and wso2 nexus since we decided not to include it in our distribution

We removed from orbit using 
https://github.com/wso2/orbit/pull/703

PR #4161, opened 2022-07-26
[Improvement] Add nashorn extension for OpenJDK Nashorn

Thumimku:

$sub

This Pr will provide a potential extension for OpenJDK Nashorn.

### Problem 1: 
How to check which engine is presented when server startup.

###  Proposed Solution 1: 
Since we don’t have a configuration, we are going to check the classForName only one time when the server starts. The current approach is to check for OpenJDK Nashorn's existence if not we are going to JDK Nashorn. 

### Problem 2: 
JS wrapper models are extending AbstractJSObject which is a class of Nashorn. (Strong couple with Nashorn engine).

###  Proposed Solution 2: 
We can create an interface to refer to these concrete classes. We can create an abstract class to hold common methods and attributes. 

Nashorn Class Diagram-Page-2

1 - Abstract JsAuthentication Context to hold common methods and attributes that implements JsBaseAuthenticationContext Interface. (Interface will be used for reference.)
2 - OpenJDK Nashorn Implementation of JsAuthenticationContext
5 - JDK Nashon Implementation of JsAuthenticationContext
4/7 - JsObject Interface from JDK Nashorn/ OpenJDK Nashorn
3/6 -  Abstract JavaScript Proxy Object for JDK Nashorn/OpenJDK Nashorn Implementation. Interface Created from modifying jdk.nashorn.api.scripting.AbstractJsObject / 
org.openjdk.nashorn.api.scripting.AbstractJsObject.

Note that AbstractJSObject will be implemented at a fine-grained level. The reason we create our own modification of AbstractJsObject is that JsNashornAuthenticationContext already extends JsAuthenticationContext.

### Problem 3: 

 In runtime, when we want to create a JS reference Obejcts (eg: JsAuthenticated User, JsAuthenticationContext, JsClaims), we don’t know which engine is active at that time.

###  Proposed Solution 3: 

We can’t check the engine type each time (less efficient). Hence we need to have Abstract Factory Model. 

Nashorn Class Diagram-Page-3

GraphBasedSequenceHandler uses JsWrapperFactoryProvider(Singleton to get the appropriate Javascript Proxy Object Wrapper Factory) to get the appropriate factory to create a Javascript Object Wrappers for JDK Nashorn/OpenJDK Nashorn execution. JsWrapperFactoryProvider will check the Nashron availability only one time and hold the appropriate Factory as an attribute.

### Problem 4: 
JsGraphBuilder (Our class with Nashorn dependency) has some Static methods, we are using these methods in other repositories too.

### Proposed Solution 4: 
We can use a Wrapper class that has similar static methods. When an external class calls these methods, the wrapper class will call internal static methods depending on the Nashorn Engine.

### Problem 5: 
JsGraphBuilder class has an inner class called JsBasedEvaluator. Since we refer to this concrete class through the interface we can’t directly call the inner class.

### Proposed Solution 5: 
Add a method to the interface to create a corresponding inner class instance. Hence we can call the method through interface reference and get the corresponding inner class instance. 

Nashorn Class DiagramPage-4

I tested role-based Adaptive authentication of the above-proposed approaches collectively and in the following scenarios and they are working as expected.

Scenario 1: JDK 11 without OpenJDKNashorn (L1)
Scenario 2: JDK 17 with OpenJDK Nashorn (L2)
Scenario 3: JDK 11 with OpenJDK Nashorn (L3)


components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityUtil.java
```diff
@@ -1284,8 +1287,8 @@ public static String convertCertificateToPEM(Certificate certificate) throws Cer
 
         byte[] encodedCertificate = org.apache.commons.codec.binary.Base64.encodeBase64(certificate.getEncoded());
 
-        String encodedPEM = String.format("%s\n%s\n%s", X509Factory.BEGIN_CERT, new String(encodedCertificate),
-                X509Factory.END_CERT);
+        String encodedPEM = String.format("%s\n%s\n%s", BEGIN_CERT, new String(encodedCertificate),
+                END_CERT);
```
senthalan, line 1291: does these changes are part of this effort ?

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/util/IdentityUtil.java
```diff
@@ -60,7 +60,7 @@
 import org.wso2.carbon.utils.CarbonUtils;
 import org.wso2.carbon.utils.NetworkUtils;
 import org.wso2.carbon.utils.multitenancy.MultitenantUtils;
-import sun.security.provider.X509Factory;
+//import sun.security.provider.X509Factory;
```
senthalan, line 63: commented import

senthalan (approved):

PR #4193, opened 2022-08-04
Fix behaviour of return_only_mapped_local_roles property

vihanga-liyanage:

We have the following property in IDP role mapping.

```
[idp_role_management]
return_only_mapped_local_roles = true
```
When this is added, the following will be the behavior.

1. If there is no IDP role mapping added, the server will return all IDP roles as they are.
2. If there is IDP role mapping and matching mapping is present, the server will send only the matched and mapped local roles.
3. If there is IDP role mapping and no matching mapping is present, the server will not send any roles.

Refers https://github.com/wso2/product-is/issues/11941


piraveena:
@vihanga-liyanage can you trigger the PR builder?

piraveena:
Build is successful
```
[INFO] Results:
[INFO] 
[INFO] Tests run: 1898, Failures: 0, Errors: 0, Skipped: 0
```

piraveena (approved):

PR #4268, opened 2022-09-14
Add getPropertyValue method for preference client

VivekVinushanth:
Improve the preferenceRetrievalClient to return propertyvalues and not just boolean preferences.


VivekVinushanth:
@DimalChandrasiri will followup on this.

PR #4286, opened 2022-09-28
Remove XACML, UMA and Workflow features from IPK

hwupathum:
- Remove identity-entitlement and policy-editor
- Remove uma
- Remove Workflow


hwupathum:
Since the PR is too big, will be added in multiple PRs

PR #4299, opened 2022-10-17
Add resolveQueryString method to AuthenticationEndpointUtils

ThaminduR:

Add `resolveQueryString` method to `AuthenticationEndpointUtils` to build the query string from the request parameter map. This is used when parameter filtering is enabled and parameters are sent in the request body instead of the URL.


mefarazath (approved):

lashinijay:
Shall we mention the related git issue in the description is there any?

PR #4309, opened 2022-10-25
Bump the version to 6.0.0 for IPK

hwupathum:


hwupathum:
Will send the PR from a separate branch

PR #4310, opened 2022-10-26
Fix authentication context not getting persisted after parameter filtering

ThaminduR:

- In the current flow, once the parameters in the redirect URL are filtered, the context object is not persisted to the cache instead framework serves it from in-memory. Authentication Context should be persisted to the cache once the parameter filtering is done.

Filtering logic is applied [here](https://github.com/wso2/carbon-identity-framework/blob/master/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultRequestCoordinator.java#L373)


mefarazath (approved):

mefarazath:
Skipping the PR builder check since this will anyways be tested during framework version bump in product is build which we plan to run immediately

PR #4330, opened 2022-11-11
Fix organization qualified request caller path issue

sadilchamishka:

As shown in the following figure, the `oauth2/authorize` request is organization qualified, and the response should be redirected to the main application owned organization's commonauth endpoint as tenant qualified request. 

Untitled Diagram-Page-9

The proposed change arise new concern if the SaaS app is owned by a sub organization. At the end of the federation the redirection should happen for the SaaS app owned sub organization and as mentioned above the org qualifier of the request and the org qualifier of the response redirection are not same, hence the redirection will happen with tenant qualified path. Should check it is even fine if that happen and more specifically SaaS app will not owned by sub organizations. (But PCC would get impacted, there fore better to check)


sadilchamishka:
The issue is fixed by this PR - https://github.com/wso2/carbon-identity-framework/pull/4372

PR #4450, opened 2023-02-16
Prevent deleting claims if the claim is not coming in the IDP

kayathiri4:
Currently, when we do custom claim mapping to OIDC IdPs, and if that attribute is not coming in the federated claims, the local claim of that federated user gets deleted (even the `attribute_sync_method` is set to `PRESERVE_LOCAL`

This PR fixes the above issue


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/claims/impl/DefaultClaimHandler.java
```diff
@@ -257,6 +258,20 @@ protected Map<String, String> handleFederatedClaims(Map<String, String> remoteCl
 
     }
 
+    /**
+     * Filter local claim mapping only if the claim is there in remote claim set.
```
sachiniWettasinghe, line 262:
```suggestion
     * Filter local claim mapping only if the claim value is there in the remote claim set.
```

sachiniWettasinghe (approved):

PR #4492, opened 2023-03-22
Add length prop to claim

kayathiri4:
$subject

### DIFF
--- a/components/claim-mgt/org.wso2.carbon.identity.claim.metadata.mgt/src/main/java/org/wso2/carbon/identity/claim/metadata/mgt/ClaimMetadataManagementServiceImpl.java
+++ b/components/claim-mgt/org.wso2.carbon.identity.claim.metadata.mgt/src/main/java/org/wso2/carbon/identity/claim/metadata/mgt/ClaimMetadataManagementServiceImpl.java
@@ -51,7 +51,24 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import static org.wso2.carbon.identity.claim.metadata.mgt.util.ClaimConstants.ErrorMessage.*;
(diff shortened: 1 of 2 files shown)


components/claim-mgt/org.wso2.carbon.identity.claim.metadata.mgt/src/main/java/org/wso2/carbon/identity/claim/metadata/mgt/ClaimMetadataManagementServiceImpl.java
```diff
@@ -522,6 +526,20 @@ private void validateClaimProperties(Map<String, String> claimProperties) throws
...
+            if (StringUtils.equalsIgnoreCase(ClaimConstants.MIN_LENGTH, property.getKey()) ||
+                    StringUtils.equalsIgnoreCase(ClaimConstants.MAX_LENGTH, property.getKey())) {
+                checkMinMaxLimit(property.getKey(), property.getValue());
+            }
+        }
+    }
+
+    private void checkMinMaxLimit(String property, String value) throws ClaimMetadataClientException {
+
+        if (!NumberUtils.isNumber(value) || Integer.parseInt(value) < MIN_CLAIM_PROPERTY_LENGTH_LIMIT ||
```
sahandilshan, line 538: Is the String value always an integer value, Don't we need to consider the cases where we get non-parsable values(non-integer values)?

kayathiri4: The expected value for these props are numbers. So we should not accept a non numeric value here.

sahandilshan (approved):

PR #4587, opened 2023-05-02
Ignore nonce cookie validation for whitelisted authenticators

sahandilshan:
* $subject
* Ignore the nonce cookie validation for whitelisted authenticators

Once this is deployed, we can configure the authenticators that we want to ignore the nonce cookie validation in the `deployement.toml`  as follows,

```
[session.nonce.cookie]
whitelist_authenticators=["Authenticator_1", "Authenticator_2"]
```

### DIFF
--- a/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
+++ b/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
@@ -1899,9 +1899,12 @@
 
     <EnableSessionNonceCookie>{{session.nonce.cookie.enabled}}</EnableSessionNonceCookie>
 
-    {% if session.nonce.cookie.whitelist_authenticators is defined %}
+    {% if session.nonce.cookie.default_whitelist_authenticators is defined or session.nonce.cookie.whitelist_authenticators is defined %}
         <SessionNonceCookie>
-            {% for whitelist_authenticator in  session.nonce.cookie.whitelist_authenticators %}
+            {% for whitelist_authenticator in session.nonce.cookie.default_whitelist_authenticators %}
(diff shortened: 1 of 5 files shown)


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -1899,6 +1899,14 @@
 
     <EnableSessionNonceCookie>{{session.nonce.cookie.enabled}}</EnableSessionNonceCookie>
 
+    {% if session.nonce.cookies.whitelist.authenticators is defined %}
```
janakamarasena, line 1902:
```suggestion
    {% if session.nonce.cookie.whitelist.authenticators is defined %}
```

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -1899,6 +1899,17 @@
 
     <EnableSessionNonceCookie>{{session.nonce.cookie.enabled}}</EnableSessionNonceCookie>
 
+    {% if session.nonce.cookie.default_whitelist_authenticators is defined %}
```
janakamarasena, line 1902: Is it possible to check if either default is defined or the custom one is defined? and not just check if the default is defined. 

Otherwise, we need to remove this condition here. Else if there are no defaults and there are user defined ones then the congif will not apply

janakamarasena (approved):

sahandilshan:
Magic Link authenticator's corss-browser support option was documented with following PRs
https://github.com/wso2/docs-is/pull/4137
https://github.com/wso2/docs-is/pull/4138

PR #4622, opened 2023-05-12
Add config to backward compatibility of filter query params from consent page url

sandushi:
### Purpose
$subject

### Changes from the PR
- The default behaviour of the consent page url is changed from https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2020
- Previously all the query parameters are appended to the consent page url.
- Now, all the query parameters are filtered from the consent page url except sessionDataKeyConsent query param by default.
- So to keep the backward compatibility, from this PR , we introduced the below config.
- If some one needs to keep the previous behavior, needs to add the below config to deployment.toml file

```
[authentication.endpoint.consent_page_redirect_params]
allow = true
```


ashensw:
@sandushi Let's open an IS migration issue also to track the behavior change.

PR #4626, opened 2023-05-15
Fix NPE in IdPManagementUIUtil

amanda-ariyaratne:

- Add null check to fix issue https://github.com/wso2/product-is/issues/15854

### DIFF
--- a/components/idp-mgt/org.wso2.carbon.idp.mgt.ui/src/main/java/org/wso2/carbon/idp/mgt/ui/util/IdPManagementUIUtil.java
+++ b/components/idp-mgt/org.wso2.carbon.idp.mgt.ui/src/main/java/org/wso2/carbon/idp/mgt/ui/util/IdPManagementUIUtil.java
@@ -303,9 +303,11 @@ public class IdPManagementUIUtil {
                     if (("googleapps").equals(provisioningConnector.getName())) {
                         Property[] googleProperties = provisioningConnector
                                 .getProvisioningProperties();
-                        for (Property property : googleProperties) {
-                            if (property.getName().equals("google_prov_private_key")) {
-                                paramMap.put("old_google_prov_private_key", property.getValue());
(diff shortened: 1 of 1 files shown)


components/idp-mgt/org.wso2.carbon.idp.mgt.ui/src/main/java/org/wso2/carbon/idp/mgt/ui/util/IdPManagementUIUtil.java
```diff
@@ -303,9 +303,11 @@ public static IdentityProvider buildFederatedIdentityProvider(HttpServletRequest
                     if (("googleapps").equals(provisioningConnector.getName())) {
                         Property[] googleProperties = provisioningConnector
                                 .getProvisioningProperties();
-                        for (Property property : googleProperties) {
-                            if (property.getName().equals("google_prov_private_key")) {
-                                paramMap.put("old_google_prov_private_key", property.getValue());
+                        if (googleProperties != null) {
+                            for (Property property : googleProperties) {
+                                if (property.getName().equals("google_prov_private_key")) {
```
chamathns, line 308: let's compare the variable to the constant. Not the constant to the variable. The reason being, in case the `property.getname()` returns a null, the `equals` method which is a method in the String class would be called on a null reference thereby throwing an NPE. Hence, as a practice, let's call the quals method on constants. 

Additionally, for instances where two variables needs to be compared, use the apache commons StringUtils.equals to avoid NPEs.

amanda-ariyaratne: Fixed with 077ff817a7fc526c8e32ebc93705e62833169910

chamathns (approved):

PR #4824, opened 2023-07-30
fix scim2 group & role issue

DilshanSenarath:

This PR will fix the issue related to the scim2 Role endpoint. Followings are the issue details.

- https://github.com/wso2/product-is/issues/15749
- https://github.com/wso2-enterprise/wso2-iam-internal/issues/459

### DIFF
--- a/components/role-mgt/org.wso2.carbon.identity.role.mgt.core/src/main/java/org/wso2/carbon/identity/role/mgt/core/dao/RoleDAOImpl.java
+++ b/components/role-mgt/org.wso2.carbon.identity.role.mgt.core/src/main/java/org/wso2/carbon/identity/role/mgt/core/dao/RoleDAOImpl.java
@@ -1449,7 +1449,6 @@ public class RoleDAOImpl implements RoleDAO {
         return roleNamesToIDs;
     }
 
-    @SuppressWarnings("checkstyle:LineLength")
     @Override
     public String getRoleNameByID(String roleID, String tenantDomain) throws IdentityRoleManagementException {
 
@@ -1481,8 +1480,7 @@ public class RoleDAOImpl implements RoleDAO {
                             + tenantDomain;
             throw new IdentityRoleManagementServerException(UNEXPECTED_SERVER_ERROR.getCode(), errorMessage, e);
         }
-        /* Verify whether the roleName is either null or a groupName
-        (excluding those with the 'Internal/Application' prefix). */
+        // Verify whether the roleName is either null or a groupName
         if (roleName == null || !SCIMCommonUtils.isHybridRole(roleName)) {
             String errorMessage = "A role doesn't exist with id: " + roleID + " in the tenantDomain: " + tenantDomain;
             throw new IdentityRoleManagementClientException(ROLE_NOT_FOUND.getCode(), errorMessage);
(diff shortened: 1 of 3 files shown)


components/role-mgt/org.wso2.carbon.identity.role.mgt.core/src/main/java/org/wso2/carbon/identity/role/mgt/core/dao/RoleDAOImpl.java
```diff
@@ -1479,7 +1480,7 @@ public String getRoleNameByID(String roleID, String tenantDomain) throws Identit
                             + tenantDomain;
             throw new IdentityRoleManagementServerException(UNEXPECTED_SERVER_ERROR.getCode(), errorMessage, e);
         }
-        if (roleName == null) {
+        if (roleName == null || !SCIMCommonUtils.isHybridRole(roleName)) {
```
DMHP, line 1483: This logic is not very clear. You are trying to throw the error message "A role doesn't exist with id: " + roleID + " in the tenant domain: " + tenant domain, if the role name is null or the role name does not start with "Application" or "Internal". But from this method, we are trying to get the role name by passing the role id to the table IDN_SCIM_GROUP. So based on the logic if we get a role that does not contain the prefix "Application" or "Internal" it will throw an error. Is this expected?

DilshanSenarath: According to the current implementation, IDN_SCIM_GROUP table stores both roles and groups. This group endpoint code (https://github.com/wso2-support/identity-inbound-provisioning-scim2/blob/645b50a897929cd5cb3d32c5a1af3b6c9c6b933c/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java#L2620-L2624) uses this method (https://github.com/wso2-support/identity-inbound-provisioning-scim2/blob/645b50a897929cd5cb3d32c5a1af3b6c9c6b933c/components/org.wso2.carbon.identity.scim2.common/src/main/java/org/wso2/carbon/identity/scim2/common/impl/SCIMUserManager.java#L2744) to differentiate between roles and groups in the same table. Since we should only get roles name (Exclude group names), we have to do a filtering. That's why I have included this logic.

Kanapriya (dismissed):

components/role-mgt/org.wso2.carbon.identity.role.mgt.core/src/main/java/org/wso2/carbon/identity/role/mgt/core/dao/RoleDAOImpl.java
```diff
@@ -1479,7 +1480,8 @@ public String getRoleNameByID(String roleID, String tenantDomain) throws Identit
                             + tenantDomain;
             throw new IdentityRoleManagementServerException(UNEXPECTED_SERVER_ERROR.getCode(), errorMessage, e);
         }
-        if (roleName == null) {
+        // Verify whether the roleName is either null or a group name
```
Kanapriya, line 1483: Please complete the comment. Its not very clear

Kanapriya, line 1483:
```suggestion
        // Verify whether the roleName is either null or it's not contain any prefix Application/Internal

Kanapriya (approved):

PR #4864, opened 2023-08-18
Fix mismatch in subject identifier

Avarjana:
### Description
Prevent modification to the subject identifier once retrieved from the DB for non federated flows as well. 

### Related issue
- https://github.com/wso2/product-is/issues/16423


dewniMW (approved):

dewniMW:
@Avarjana let's make sure to document this change too

PR #4889, opened 2023-09-01
[Bug-fix] Fix JSON exception bug when rules not available

Thumimku:
When rules were not available in the configuration (ex: in a fresh pack only regex validator was available), 
```
config.get("rules");
```
throws JSON exception.

Fix: Add a has check before get rules json object


madurangasiriwardena (approved):

kayathiri4 (approved):

DilshanSenarath:
I have successfully tested this fix in my local setup, and it performed as expected.

PR #4895, opened 2023-09-04
Remove keystore module.

mpmadhavig:
### Describe
This effort will remove the keystore package from framework security-mgt module and move that to the kernel. The decision has been taken to do so was due to a cyclic dependency between the keystore package and the keyStoreAdmin in carbon-core.

### Related issues
- https://github.com/wso2/product-is/issues/16422


madurangasiriwardena:
To minimize the external impact from this change, instead of removing the OSGi services and SOAP services, we can update the implementation to invoke the new service you have introduced in the kernel. And we can deprecate these services as well.

PR #4906, opened 2023-09-11
Adding FAPI related OpenID configurations

ChinthakaJ98:
## Related Issues
https://github.com/wso2/product-is/issues/16924


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -839,6 +839,21 @@
...
             </FAPI>
             {% endif %}
+
+            <TLSClientCertificateBoundAccessTokensSupported>{{oauth.oidc.tls_client_certificate_bound_access_tokens_supported}}</TLSClientCertificateBoundAccessTokensSupported>
+
+            <SupportedTokenEndpointAuthMethods>
+                {% for authmethod in oauth.oidc.token_endpoint.auth_methods %}
+                    <SupportedTokenEndpointAuthMethod>{{authmethod}}</SupportedTokenEndpointAuthMethod>
+                {% endfor %}
+            </SupportedTokenEndpointAuthMethods>
```
janakamarasena, line 849: These two should not be server configs but rather resolved from the components running in the system. 
IIRC we don't have any configs for adding the actual components to the system either.

ChinthakaJ98: Fixed in 9078393eebc9a741942a9a4cd60f477abf103517

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json
```diff
@@ -234,6 +234,8 @@
   "oauth.oidc.request_object_builder.request_param_value_builder.enabled": true,
   "oauth.oidc.request_object_builder.request_param_value_builder.class": "org.wso2.carbon.identity.openidconnect.RequestParamRequestObjectBuilder",
   "oauth.oidc.redirect_to_post_logout_uri_on_consent_denial": true,
+  
+  "oauth.oidc.token_endpoint.signing_algorithms": ["PS256", "ES256"],
```
janakamarasena, line 238: Is the currently used algorithm also added to this? Also in what places is this signing algs used?

SachiniSiriwardene: is it okay to make this as a common configuration to fetch the supported signing algorithms in general?

janakamarasena: are we showing the token ep supported singing algs in the discovery document? if so we should show all server supported algs(IINM currently we pick one specific alg so we can add that one here). If there is no requirement to maintain such a common config lets make this config fapi specific.

ChinthakaJ98: We need to show the token ep. supported signing algorithms in the discovery. And in that list we need to have atleast one from PS256 and ES256 along with the other algorithms that the server supports. 

Here the concern is as the algorithms will be validated in DCR, whether we could have a common config with all the server supported algorithms without making it specific saying token endpoint. For the discovery, we can use that same config as well without an issue. Is it alright?

janakamarasena: in actuality are all the algs usable in all the places?

ChinthakaJ98: Kept the configuration specific to the token endpoint as discussed offline. Also did include the currently used algorithm as well to the default.json in d04a248cb24fbbaa1c213264dd6babced00ccdbc

janakamarasena (approved):

PR #4946, opened 2023-09-25
Adding required configurations for FAPI validations

ChinthakaJ98:
```
[oauth.oidc.fapi]
allowed_client_authentication_methods=["private_key_jwt","tls_client_auth"]
allowed_signature_algorithms=["PS256","ES256"]
```

## Related Issues
https://github.com/wso2/product-is/issues/16519


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -829,16 +829,34 @@
...
             -->
-            {% if oauth.oidc.fapi is defined %}
             <FAPI>
                 {% if oauth.oidc.fapi.enable_ciba_profile is defined %}
                 <EnableCibaProfile>{{oauth.oidc.fapi.enable_ciba_profile}}</EnableCibaProfile>
                 {% endif %}
                 {% if oauth.oidc.fapi.enable_security_profile is defined %}
                 <EnableSecurityProfile>{{oauth.oidc.fapi.enable_security_profile}}</EnableSecurityProfile>
                 {% endif %}
+                <SupportedClientAuthenticationMethods>
```
janakamarasena, line 839: Instead of the term supported lets use allowed.
```suggestion
                <AllowedClientAuthenticationMethods>
```

janakamarasena:
Can you add the sample deployment.toml config to the PR description.

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -829,16 +829,34 @@
...
+                    {% if oauth.oidc.fapi.client_authentication_methods|length %}
+                        {% for authmethod in oauth.oidc.fapi.client_authentication_methods %}
+                            <SupportedClientAuthenticationMethod>{{authmethod}}</SupportedClientAuthenticationMethod>
+                        {% endfor %}
+                    {% else %}
```
janakamarasena, line 844: Instead of having a else if we can add the defaults to the default.json so this conditional check is auto-handled. If no configs are defined the default.json config will apply here.

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -829,16 +829,34 @@
...
+                        <SupportedClientAuthenticationMethod>private_key_jwt</SupportedClientAuthenticationMethod>
+                        <SupportedClientAuthenticationMethod>tls_client_auth</SupportedClientAuthenticationMethod>
+                    {% endif %}
+                </SupportedClientAuthenticationMethods>
+                <SupportedSignatureAlgorithms>
```
janakamarasena, line 849: In what places are these signature algorithms used?

ChinthakaJ98: This is used at the token endpoint when validating the signature algorithm. If the algorithm is not specified in the app level, this is the fallback option. 
https://github.com/wso2-extensions/identity-oauth-addons/pull/108/files#diff-c1fefc87f2cf0485b2f5e352367c515cb1036459a5232698108ff48d4d50a5e1R760

@SachiniSiriwardene this will be used for DCR validations as well right?

janakamarasena: shall we make this config name specific to the place where we are using this (ex: token_endpoint_auth_signing_alg,  request_object_signing_alg , etc). Right now when we say supported it has broad usage.

ChinthakaJ98: We made the name common and nested it under the <FAPI> parent tag because the spec mentions the same set of algorithms for both token endpoint and request object signing [1]. Therefore shall we keep the name as it is or do you feel that we need to change the config names based on the usage? WDYT?

[1] https://openid.net/specs/openid-financial-api-part-2-1_0.html#algorithm-considerations

janakamarasena: even though the spec mentions a common set of algorithms when an app registers them it is registered under different attributes right? And the usage is in specific unrelated cases, therefore I think it will be cleaner to maintain these separately.

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
ChinthakaJ98: Fixed in 572230c036c602e133e13c0aeb2ab88f7e3a74f1

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
ChinthakaJ98: Fixed in 572230c036c602e133e13c0aeb2ab88f7e3a74f1

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
SachiniSiriwardene: but still for validating we just need to check whether the algorithm is one of the above. It doesn't matter which use case it is (token assertion/request object). Those attributes will be stored separately. These values are just used to filter whether the sent attributes are one of the allowed ones for a FAPI application.

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -829,16 +829,24 @@
...
-            {% if oauth.oidc.fapi is defined %}
             <FAPI>
                 {% if oauth.oidc.fapi.enable_ciba_profile is defined %}
                 <EnableCibaProfile>{{oauth.oidc.fapi.enable_ciba_profile}}</EnableCibaProfile>
                 {% endif %}
                 {% if oauth.oidc.fapi.enable_security_profile is defined %}
                 <EnableSecurityProfile>{{oauth.oidc.fapi.enable_security_profile}}</EnableSecurityProfile>
                 {% endif %}
+                <AllowedClientAuthenticationMethods>
+                    {% for authmethod in oauth.oidc.fapi.client_authentication_methods %}
```
janakamarasena, line 840:
```suggestion
                    {% for authmethod in oauth.oidc.fapi.allowed_client_authentication_methods %}
```

janakamarasena (approved):

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -829,16 +829,24 @@
...
+                        <AllowedClientAuthenticationMethod>{{authmethod}}</AllowedClientAuthenticationMethod>
+                    {% endfor %}
+                </AllowedClientAuthenticationMethods>
+                <AllowedSignatureAlgorithms>
+                    {% for algorithm in oauth.oidc.fapi.signature_algorithms %}
```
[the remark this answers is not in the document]
ChinthakaJ98: Fixed in fc1aa0a6a67deeefd4c1877b09994765b8f7a37f and 026691ef75ba47fd509141c4c54ca4afd678d2ed

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
ChinthakaJ98: Fixed in fc1aa0a6a67deeefd4c1877b09994765b8f7a37f and 026691ef75ba47fd509141c4c54ca4afd678d2ed

PR #4971, opened 2023-10-05
Skip set commonauth cookie

VivekVinushanth:
- Partly supports: https://github.com/wso2/product-is/issues/16855
-  In this PR, we provide a way to skip setting commonauth cookie to the response.
- Since we are extending password grant to utilize authentication framework, while it creates a session , it sets a common-auth cookie back in the response. But for password grant there is no necessity for session. Though we force authentication again, it is advisable to not return commonauth cookie since the session can be manipulated from a different device.

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultAuthenticationRequestHandler.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultAuthenticationRequestHandler.java
@@ -623,7 +623,7 @@ public class DefaultAuthenticationRequestHandler implements AuthenticationReques
                         authenticationResult.getSubject().getUserName(), sessionContextKey,
                         authenticationResult.getSubject().getTenantDomain(), FrameworkUtils.getCorrelation(),
                         createdTimeMillis, sessionContext.isRememberMe());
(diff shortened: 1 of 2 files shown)


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultAuthenticationRequestHandler.java
```diff
@@ -622,7 +623,10 @@ protected void concludeFlow(HttpServletRequest request, HttpServletResponse resp
                         authenticationResult.getSubject().getUserName(), sessionContextKey,
                         authenticationResult.getSubject().getTenantDomain(), FrameworkUtils.getCorrelation(),
                         createdTimeMillis, sessionContext.isRememberMe());
-                setAuthCookie(request, response, context, sessionKey, applicationTenantDomain);
+                if (request.getAttribute(SKIP_SET_COMMONAUTH_COOKIE) == null
```
pulasthi7, line 626: shall we use something like skipSessionCreation, or skipWebSession? which will make more sense to outsider, and will be more generic and can be used to skip any other such web session specific things

pulasthi7: also, let's have the positive form in configs to avoid double negation.

pulasthi7 (approved):

PR #4974, opened 2023-10-06
Update the SessionContextCache after getting from DB

hwupathum:
Fix https://github.com/wso2/product-is/issues/16819

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/cache/SessionContextCache.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/cache/SessionContextCache.java
@@ -178,6 +178,12 @@ public class SessionContextCache extends BaseCache<SessionContextCacheKey, Sessi
                     return null;
                 }
             }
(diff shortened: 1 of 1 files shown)


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/cache/SessionContextCache.java
```diff
@@ -178,6 +178,12 @@ private SessionContextCacheEntry getSessionFromDB(SessionContextCacheKey key) {
                     return null;
                 }
             }
+            // Add to cache if session is available in the database.
+            if (log.isDebugEnabled()) {
+                log.debug("Session corresponding to the key : " + key.getContextId() +
+                        " found in the db and is added to the cache.");
+            }
+            super.addToCache(key, cacheEntry, resolveLoginTenantDomain(getLoginTenantDomainFromContext()));
```
JKAUSHALYA, line 186: Shall we add a comment here explaining why we are adding to the cache immediately. Because later in the flow, we update the values and re add to the cache. That might confuse developers because we modify the same value and makes updating here redundant.

JKAUSHALYA (approved):

PR #5013, opened 2023-10-17
Add method to get configuration resource from tenant ID

dewniMW:
When getting the resource from ConfigurationManager currently the tenant id is fetched from the context. However, there are cases such as in organization discovery where we need to fetch the resource by passing the tenant id (instead of getting it directly from the context). Hence introducing a new method to achieve this.

### DIFF
--- a/components/configuration-mgt/org.wso2.carbon.identity.configuration.mgt.core/src/main/java/org/wso2/carbon/identity/configuration/mgt/core/ConfigurationManager.java
+++ b/components/configuration-mgt/org.wso2.carbon.identity.configuration.mgt.core/src/main/java/org/wso2/carbon/identity/configuration/mgt/core/ConfigurationManager.java
@@ -167,6 +167,18 @@ public interface ConfigurationManager {
     Resource getResource(String resourceTypeName, String resourceName) throws ConfigurationManagementException;
 
     /**
+     * This API is used to retrieve the given resource by the tenant ID.
+     *
+     * @param tenantId         The ID of the tenant.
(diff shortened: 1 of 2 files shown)


thanujalk (approved):

components/configuration-mgt/org.wso2.carbon.identity.configuration.mgt.core/src/main/java/org/wso2/carbon/identity/configuration/mgt/core/ConfigurationManager.java
```diff
@@ -166,6 +166,18 @@ default Resource replaceResource(String resourceTypeName, Resource resource)
...
+    /**
+     * This API is used to retrieve the given resource by the tenant ID.
+     *
+     * @param tenantId         The ID of the tenant.
+     * @param resourceTypeName The name of the {@link ResourceType}.
+     * @param resourceName     The name of the {@link ResourceType}.
+     * @return {@link Resource}
+     * @throws ConfigurationManagementException Configuration management exception.
+     */
+    Resource getResourceByTenantId(int tenantId, String resourceTypeName, String resourceName) throws
```
AnuradhaSK, line 178: this should be a default method, since we are adding new service to an existing interface

dewniMW: fixed with https://github.com/wso2/carbon-identity-framework/pull/5015

PR #5048, opened 2023-10-24
Remove unwanted adding of super tenant to caller path

ImalshaG:
This PR adds the super tenant with tenant prefix to the caller path when building the commonauth url according to the `isSuperTenantRequiredInUrl` config.

Related issue: https://github.com/wso2/product-is/issues/16906


Thisara-Welmilla:
It is decided to not do this inbound restriction, hence do not required this.

PR #5207, opened 2023-11-23
Fix IDP metadata retrieval issue and recovery endpoint service URL builder issue

sadilchamishka:

There are two fixes sent with this PR
1 - There are internal API calls which are invoked from the preference client to retrieve metadata information. When accessing organization resources, the service URLs should be in the format of `/t/<tenant-domain>/o/`. But the retrieval clients used the product is not following client-credentials grant type based token used approach, instead a basic auth based approach. Hence the mentioned `/t/<tenant-domain>/o/` can't be followed. Hence for this case only, call the organization resource APIs with `/t/<tenant-domain>` path, where the tenant domain is the corresponding tenant of the organization. Here the organization-id & tenant-domain are assumed to be equal when considering organizations.

The above discussed API calls are invoked internally, until properly fix the retrieval clients with proper authentication mechanisms, this fix has to be applied.

2 - There was an issue in service URL building for the recovery endpoint in the ask password feature of the organization level.

### Related Issues
1 
- https://github.com/wso2/product-is/issues/17591

2 
- https://github.com/wso2/product-is/issues/18057

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/ConfigurationFacade.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/ConfigurationFacade.java
@@ -356,7 +356,8 @@ public class ConfigurationFacade {
             }
         }
         try {
-            return ServiceURLBuilder.create().addPath(path).build().getAbsolutePublicURL();
+            String organizationId = PrivilegedCarbonContext.getThreadLocalCarbonContext().getOrganizationId();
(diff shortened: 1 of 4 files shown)


thanujalk (approved):

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/config/ConfigurationFacade.java
```diff
@@ -356,7 +356,9 @@ private String buildAbsoluteUrl(String defaultContext, Supplier<String> getValue
             }
```
sadilchamishka, line 356: This method is called for building authentication and recovery endpoint URLs.
Screenshot 2023-11-24 at 10 05 03

dewniMW (approved):

PR #5325, opened 2023-12-20
Get application access URL for tenants

JeethJJ:
If the server is not running in legacy mode, we can retrieve the application access URL for the specific tenant itself since it's not a saas application in the new runtime. 

Issue : https://github.com/wso2/product-is/issues/18430

### DIFF
--- a/components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint.util/src/main/java/org/wso2/carbon/identity/mgt/endpoint/util/IdentityManagementEndpointUtil.java
+++ b/components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint.util/src/main/java/org/wso2/carbon/identity/mgt/endpoint/util/IdentityManagementEndpointUtil.java
@@ -34,6 +34,7 @@ import org.apache.cxf.jaxrs.client.JAXRSClientFactoryBean;
 import org.json.JSONArray;
 import org.json.JSONObject;
 import org.owasp.encoder.Encode;
+import org.wso2.carbon.CarbonConstants;
 import org.wso2.carbon.base.MultitenantConstants;
 import org.wso2.carbon.core.SameSiteCookie;
 import org.wso2.carbon.identity.application.authentication.framework.util.FrameworkConstants;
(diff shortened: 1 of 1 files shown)


components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint.util/src/main/java/org/wso2/carbon/identity/mgt/endpoint/util/IdentityManagementEndpointUtil.java
```diff
@@ -188,11 +189,17 @@ public static final String getUserPortalUrl(String userPortalUrl, String tenantD
...
-                            My_ACCOUNT_APPLICATION_NAME);
+                    if (CarbonConstants.ENABLE_LEGACY_AUTHZ_RUNTIME == true) {
+                        myAccountAccessUrl = applicationDataRetrievalClient.getApplicationAccessURL(SUPER_TENANT,
+                                My_ACCOUNT_APPLICATION_NAME);
+                    } else {
+                        myAccountAccessUrl = applicationDataRetrievalClient.getApplicationAccessURL(tenantDomain,
+                                My_ACCOUNT_APPLICATION_NAME);
+                    }
                     if (StringUtils.isNotEmpty(myAccountAccessUrl)) {
-                        return replaceUserTenantHintPlaceholder(myAccountAccessUrl, tenantDomain);
```
Thisara-Welmilla, line 195: Don't we need to do this when `CarbonConstants.ENABLE_LEGACY_AUTHZ_RUNTIME == true`?

JeethJJ: If legacy runtime is enabled, we will be using the SAAS app. So the method to log into a tenanted user in the SAAS app is to provide the username as `name@tenantDomain`. So, it is not required.

Thisara-Welmilla: But if the url contains the `USER_TENANT_HINT_PLACE_HOLDER` placeholder by any change, don't we need to replace with proper values.

Thisara-Welmilla (approved):

components/identity-mgt/org.wso2.carbon.identity.mgt.endpoint.util/src/main/java/org/wso2/carbon/identity/mgt/endpoint/util/IdentityManagementEndpointUtil.java
JeethJJ: As I tested, `USER_TENANT_HINT_PLACE_HOLDER` was not present when we retrieve the super tenant myaccount access url in legacy mode. 

And in default runtime, we directly retrieve the access url for the specific tenant. Hence, this is not required in either.

PR #5539, opened 2024-02-22
Fix idle session timeout for a tenant being taken from super tenant

AmshikaH:

- Addresses https://github.com/wso2/product-is/issues/5115#issuecomment-968301066.


AmshikaH:
Previously, the getSessionContextFromCache method required starting the tenant flow prior to calling this method in order to ensure the correct tenant domain was retrieved. However, the previous method was [deprecated](https://github.com/wso2/carbon-identity-framework/blob/689acb30fdb895a1a2223829c8d71366ad6819bc/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/util/FrameworkUtils.java#L1215) and the new getSessionContextFromCache currently in use does not have this issue. 

Therefore, this fix is not necessary.

PR #5612, opened 2024-04-01
Check if authenticated authenticator is not null during logout flow

bhagyasakalanka:
Check if authenticated authenticator is not null during logout flow

This fix addresses the logout issue arising when the prompt adaptive function is utilized for user authentication following an external redirection. The challenge stems from the HttpServletRequest object being newly instantiated after such redirections. Consequently, authentication initiated in the onSuccess event handler of prompt can lead to the loss of previously set flow status in the request object. This loss may result in steps lacking authenticated authenticators, thus causing logout failures.

To resolve this, the fix entails ignoring the processing of steps lacking authenticated authenticators during the logout flow.

Related PRs: https://github.com/wso2-extensions/identity-local-auth-basicauth/pull/200

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultLogoutRequestHandler.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultLogoutRequestHandler.java
@@ -256,7 +256,11 @@ public class DefaultLogoutRequestHandler implements LogoutRequestHandler {
                 if (authenticatorConfig == null) {
                     authenticatorConfig = sequenceConfig.getAuthenticatedReqPathAuthenticator();
                 }
-                // If there is no authenticated authenticator, skip the step.
+                /*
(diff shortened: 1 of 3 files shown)


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultLogoutRequestHandler.java
```diff
@@ -256,6 +256,12 @@ public void handle(HttpServletRequest request, HttpServletResponse response, Aut
                 if (authenticatorConfig == null) {
                     authenticatorConfig = sequenceConfig.getAuthenticatedReqPathAuthenticator();
                 }
+                // If there is no authenticated authenticator, skip the step.
+                if (authenticatorConfig == null) {
```
janakamarasena, line 260: Introduce a unit test covering this case.

bhagyasakalanka: fixed

janakamarasena (approved):

PR #5763, opened 2024-06-26
Implement integration templates packaging process

DilshanSenarath:
## Purpose
Previously, application templates were part of the extension management feature itself. With the SSO template onboarding feature, we have moved the application templates to a separate repository (https://github.com/wso2-extensions/identity-integration-ui-templates). In this new setup, each template is released separately, with specific versions, and published as GitHub releases.

This PR introduces logic to download the specified release in `integration-templates.xml` and include those templates into the extension management feature.

## Related Issues

- https://github.com/wso2/product-is/issues/20107


pom.xml
```diff
@@ -1808,6 +1808,13 @@
                 <artifactId>icu4j</artifactId>
                 <version>${icu.version}</version>
             </dependency>
+
+            <!--Ant-contrib for the extended functionality within the Ant build process-->
+            <dependency>
+                <groupId>ant-contrib</groupId>
+                <artifactId>ant-contrib</artifactId>
+                <version>${version.ant.contrib}</version>
+            </dependency>
```
janakamarasena, line 1817: This is not needed at runtime right? lets scope this dependency to compile time

PR #5854, opened 2024-08-12
Improve auth framework with authenticator type.

Thisara-Welmilla:
Task issue:
- https://github.com/wso2/product-is/issues/20887

With above feature, we will introduce an authentication adapter that extends the `AbstractApplicationAuthenticator` and implements the `FederatedApplicationAuthenticator` class. This adapter will support to authenticated both LOCAL and FEDERATED users and user type will be defined at runtime. The authentication flow will need to accommodate both types. To achieve this, the following improvements will introduced with this PR:

1. Introduced new authenticator types (LOCAL, FEDERATED, REQUEST_PATH, FLOW_HANDLER, CUSTOM, UNDEFINED).
2. Introduce new method, `getAuthenticatorType()` to the `ApplicationAuthenticator` interface, with UNDEFINED as the default return value.
3. The authenticator interfaces that extend `ApplicationAuthenticator` will override the `getAuthenticatorType()` method to return their specific authenticator type.
4. The some authentication flows logic differs based on the `instanceof FederatedApplicationAuthenticator` check, which ideally should be based on the authenticated user (`isFedUser()`) of the corresponding step. Those will be replaced with `AuthenticatorType.FEDERATED.equals(authenticator.getAuthenticatorType())` to support the existing behavior.
5. To handle user type for the custom authentication adapter at above(4) places, the following check will be added: `(AuthenticatorType.CUSTOM.equals(authenticator.getAuthenticatorType()) && stepConfig.getAuthenticatedUser().isFederatedUser())`.


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/step/impl/DefaultStepHandler.java
```diff
@@ -713,7 +713,8 @@ protected void doAuthentication(HttpServletRequest request, HttpServletResponse
         }
 
         String idpName = FrameworkConstants.LOCAL_IDP_NAME;
-        if (context.getExternalIdP() != null && authenticator instanceof FederatedApplicationAuthenticator) {
+        if (context.getExternalIdP() != null && (AuthenticatorType.FEDERATED.equals(authenticator
+                .getAuthenticatorType()) || AuthenticatorType.CUSTOM.equals(authenticator.getAuthenticatorType()))) {
```
malithie, line 717: Don't we need to check for the user here ?

Thisara-Welmilla: At this point, the authentication have not happened yet, therefore, cannot check user type. Also this check is just  to decide, what should be goes to diagnostic log; whether LOCAL and specific idp name based on the authenticator Type (as local authenticators does not have associated IDP).

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/AuthenticationFlowHandler.java
```diff
@@ -23,4 +23,10 @@
  */
 public interface AuthenticationFlowHandler extends ApplicationAuthenticator {
 
+    @Override
+    default AuthenticatorType getAuthenticatorType() {
+
+        return AuthenticatorType.FLOW_HANDLER;
+    }
+
```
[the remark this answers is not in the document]
Thisara-Welmilla: No this authenticator get returned back in authenticator API. But as we are overriding this method in all other marker interface, though of adding for this also.

If this is unnecessary will remove this.

PR #5864, opened 2024-08-15
Remove unnecessary dependency to Identity Util to read action execution configs

malithie:

Resolves https://github.com/wso2/product-is/issues/20915


ashanthamara:
Shall we set this config to `true` since we are enabling the feature by default. As the Actions UI is also there in the console, user might get confused if he missed to set this config to true.
https://github.com/wso2/carbon-identity-framework/blob/19b6f36bcc87c1fea8105c6b7b469f11a0a64146/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json#L1570

And do we need to mention about this config in Identity Server docs ?

ashanthamara (approved):

malithie:
> Shall we set this config to `true` since we are enabling the feature by default. As the Actions UI is also there in the console, user might get confused if he missed to set this config to true.
> 
> https://github.com/wso2/carbon-identity-framework/blob/19b6f36bcc87c1fea8105c6b7b469f11a0a64146/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json#L1570
> 
> And do we need to mention about this config in Identity Server docs ?
This is a feature flag. Therefore, it makes sense to add it if we are having a config catalog. But looks like we don't have such. In that case IMO we don't need to mention this if the feature related doc

PR #5877, opened 2024-08-21
Fix reading header when only one header is configured.

malithie:

When only one header or parameter is configured in the server level configuration to be excluded in actions request it gets rather skipped as per the current implementation.
This PR fixes that issue.

Relates to https://github.com/wso2/product-is/issues/20926 which incorporated this behaviour

soon

### DIFF
--- a/components/action-mgt/org.wso2.carbon.identity.action.execution/src/main/java/org/wso2/carbon/identity/action/execution/util/ActionExecutorConfig.java
+++ b/components/action-mgt/org.wso2.carbon.identity.action.execution/src/main/java/org/wso2/carbon/identity/action/execution/util/ActionExecutorConfig.java
@@ -154,6 +154,10 @@ public class ActionExecutorConfig {
             return Collections.emptyList();
         }
 
+        if (propertyValue instanceof String) {
+            return Collections.singletonList(propertyValue.toString());
+        }
+
         if (propertyValue instanceof List) {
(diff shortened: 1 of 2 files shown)


components/action-mgt/org.wso2.carbon.identity.action.execution/src/main/java/org/wso2/carbon/identity/action/execution/util/ActionExecutorConfig.java
```diff
@@ -154,6 +154,10 @@ private List<String> getPropertyValues(String propertyKey) {
             return Collections.emptyList();
         }
 
+        if (propertyValue instanceof String) {
+            return Collections.singletonList(propertyValue.toString());
+        }
+
```
Thisara-Welmilla, line 160: Is there reason why should be allow to configure single value as String instead of list of one element.

Thisara-Welmilla (approved):

components/action-mgt/org.wso2.carbon.identity.action.execution/src/main/java/org/wso2/carbon/identity/action/execution/util/ActionExecutorConfig.java
malithie: The API contract is to get a list of property values. So we are returning a list adhering to that.

PR #5901, opened 2024-08-31
Add authentication Types for authenticators.

Thisara-Welmilla:
Task issue:
- https://github.com/wso2/product-is/issues/20887

With above feature, we will introduce an authentication adapter that extends the `AbstractApplicationAuthenticator` and implements the `FederatedApplicationAuthenticator` class. This adapter will support to authenticated both LOCAL and FEDERATED users and user type will be defined at runtime. The authentication flow will need to accommodate both types. To achieve this, the following improvements will introduced with this PR:

1. Introduced new authenticator types (SYSTEM and CUSTOM).
2. Introduce new method, `getAuthenticatorType()` to the `ApplicationAuthenticator` interface, with SYSTEM as the default return value.
3. The authenticator extension will override the `getAuthenticatorType()` method to return their CUSTOM type.
4. The some authentication flows logic differs based on the `instanceof FederatedApplicationAuthenticator` check, which ideally should be based on the authenticated user (`isFedUser()`) of the corresponding step. Those will be updated.

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/JITProvisioningPostAuthenticationHandler.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/JITProvisioningPostAuthenticationHandler.java
@@ -34,6 +34,7 @@ import org.wso2.carbon.consent.mgt.core.model.ReceiptInput;
 import org.wso2.carbon.consent.mgt.core.model.ReceiptPurposeInput;
 import org.wso2.carbon.consent.mgt.core.model.ReceiptServiceInput;
 import org.wso2.carbon.identity.application.authentication.framework.ApplicationAuthenticator;
+import org.wso2.carbon.identity.application.authentication.framework.ApplicationAuthenticator.AuthenticatorType;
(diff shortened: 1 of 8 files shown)


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/JITProvisioningPostAuthenticationHandler.java
```diff
@@ -187,7 +188,11 @@ private PostAuthnHandlerFlowStatus handleResponseFlow(HttpServletRequest request
             AuthenticatorConfig authenticatorConfig = stepConfig.getAuthenticatedAutenticator();
             ApplicationAuthenticator authenticator = authenticatorConfig.getApplicationAuthenticator();
 
-            if (authenticator instanceof FederatedApplicationAuthenticator) {
+            AuthenticatorType authenticatorType = authenticator.getAuthenticatorType();
+            if ((AuthenticatorType.SYSTEM.equals(authenticatorType) && authenticator instanceof
+                    FederatedApplicationAuthenticator) || (AuthenticatorType.CUSTOM.equals(authenticatorType) 
+                    && stepConfig.getAuthenticatedUser().isFederatedUser())) {
```
malithie, line 194: does step config has the respective user authenticated by the respective authenticator ?

malithie:
Let's cover the test cases along with the PR or let's merge to a feature branch and merge as tests are covered only

PR #5915, opened 2024-09-05
Role based user provisioning improvements

sadilchamishka:

$subject

### Related Issues
- https://github.com/wso2/product-is/issues/11394


components/provisioning/org.wso2.carbon.identity.provisioning/src/main/java/org/wso2/carbon/identity/provisioning/listener/ProvisioningRoleMgtListener.java
```diff
@@ -0,0 +1,124 @@
...
+
+            String[] newUserNames = resolveUserNameFromUserIds(newUserIDList, userStoreManager);
+
+            if (deletedUserNames.length > 0 && newUserNames.length > 0) {
+                return;
+            }
+
+            RoleBasicInfo roleBasicInfo = ProvisioningServiceDataHolder.getInstance().getRoleManagementService()
+                    .getRoleBasicInfoById(roleId,tenantDomain);
+            // Only organization audience roles are supported for role based outbound provisioning
```
IsuraD, line 99: Shouldn't this be backwards compatible?

sadilchamishka: Before IS 7.0 all the roles are organization audience roles. Hence previous roles and new roles created with organization audience role can be used for role base outbound provisioning.

The new type of role which is application audience roles came with IS 7.0. How to handle the role based provisioning for those types of roles should be analyzed due to below concerns.
- Role name are not unique hence we have to register role uuid or role prefixed with its corresponding app name
- As outbound provisioning is organization wide config by default (we can configure app wise also), the use cases of having application roles for role based outbound provisioning required to be analyzed.

Due to above concerns, outbound provisioning for application audience is not given at first phase. 

Hope the explanation is clear enough, please raise if there any concerns. Here backward compatibility is preserved as here the application roles means not the `Application/<role-name>` role type had in previous IS versions.

chamathns (approved):

PR #5939, opened 2024-09-18
Add ability to switch SAML storage type via config

Osara-B:

- Ability to set SAML data storage type to database, registry and hybrid mode via config is implemented
- Default is set to database usage

### DIFF
--- a/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
+++ b/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
@@ -87,6 +87,7 @@
         <NotificationTemplates>{{data_storage_type.notification_templates}}</NotificationTemplates>
         <KeyStores>{{data_storage_type.keystores}}</KeyStores>
         <XACML>{{data_storage_type.xacml}}</XACML>
+        <SAML>{{data_storage_type.saml}}</SAML>
     </DataStorageType>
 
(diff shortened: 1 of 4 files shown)


darshanasbg (approved):

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -87,6 +87,7 @@
         <NotificationTemplates>{{data_storage_type.notification_templates}}</NotificationTemplates>
         <KeyStores>{{data_storage_type.keystores}}</KeyStores>
         <XACML>{{data_storage_type.xacml}}</XACML>
+        <SAML>{{data_storage_type.saml}}</SAML>
```
ImalshaG, line 90: Don't we need to add this to default.json as well to set database as default value?

Osara-B: https://github.com/wso2/carbon-identity-framework/blob/d6c2b0df87e11340ffcab6a88430a7b4dc794d03/components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/SAMLSSOServiceProviderManager.java#L55C9-L55C107 makes sure that by default the database will be used unless registry or hybrid type is specified.

ImalshaG (approved):

PR #5995, opened 2024-10-07
Add spId query param to non organization authenticator auth redirects

RushanNanayakkara:

The MFA authentication pages does not applying application even though application level branding is enabled. This is due to the authenticator page URL not having the `spId` parameter. 
This PR adds the spId parameter to all authentication redirect urls that are run through the default request coordinator if they are not being processed through `OrganizationAuthenticator`.  This is since when OrganizationAuthenticator handles the request, it's at the point of SSO ing to sub org. At this point the context still holds the main organization applications `spId'. Hence adding that `spId` will result in having main org app branding in the sub org application. This scenario is already handled separately.

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultRequestCoordinator.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultRequestCoordinator.java
@@ -75,9 +75,11 @@ import org.wso2.carbon.utils.DiagnosticLog;
 import java.io.IOException;
 import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
+import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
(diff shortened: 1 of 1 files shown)


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultRequestCoordinator.java
```diff
@@ -393,6 +396,23 @@ && isStepHasMultiOption(context)
...
+                                redirectURL = redirectURL + "&" + FrameworkConstants.REQUEST_PARAM_SP_UUID
+                                        + "=" + URLEncoder.encode(context.getServiceProviderResourceId(),
+                                        StandardCharsets.UTF_8.name());
+                            }
+                        } else {
+                            redirectURL = redirectURL + "?" + FrameworkConstants.REQUEST_PARAM_SP_UUID
+                                    + "=" + URLEncoder.encode(context.getServiceProviderResourceId(),
+                                    StandardCharsets.UTF_8.name());
+                        }
+                        responseWrapper.sendRedirect(redirectURL);
```
bhagyasakalanka, line 414: is it correct to do a redirection here?

RushanNanayakkara: it's not redirecting. It's only setting the redirect url in the responseWrapper object. The function name is mileading.

bhagyasakalanka (approved):

PR #6021, opened 2024-10-13
Fix branding not applied to mfa

RushanNanayakkara:

The MFA authentication pages does not applying application even though application level branding is enabled. This is due to the authenticator page URL not having the `spId` parameter. 
This PR adds the spId parameter to all authentication redirect urls that are run through the default request coordinator if they are not being processed through `OrganizationAuthenticator`.  This is since when OrganizationAuthenticator handles the request, it's at the point of SSO ing to sub org. At this point the context still holds the main organization applications `spId'. Hence adding that `spId` will result in having main org app branding in the sub org application. This scenario is already handled separately.

### Related Issue 
- https://github.com/wso2/product-is/issues/21314

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultRequestCoordinator.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultRequestCoordinator.java
@@ -75,9 +75,11 @@ import org.wso2.carbon.utils.DiagnosticLog;
 import java.io.IOException;
 import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
+import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URLDecoder;
 import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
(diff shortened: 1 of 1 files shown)


Malith-19 (approved):

shashimalcse (approved):

Avarjana (approved):

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/request/impl/DefaultRequestCoordinator.java
```diff
@@ -393,6 +396,10 @@ && isStepHasMultiOption(context)
 
                 if (!context.isLogoutRequest()) {
                     FrameworkUtils.getAuthenticationRequestHandler().handle(request, responseWrapper, context);
+                    // Adding spIp param to the redirect URL if the authenticator is not the organization authenticator.
+                    if (!ORGANIZATION_AUTHENTICATOR.equals(request.getParameter(AUTHENTICATOR))) {
```
sadilchamishka, line 400: Here we are trying skip adding sp-uuid to the consent page when it is organization authenticator. Trying to understand the how it created an issue. I hope you can proceed with the improvement exclude this and explain the issue may be via sharing the network trace where we can identify how it caused an issue.

RushanNanayakkara: this is not for consent page. It's for the MFA pages in sub org applications. 
When organization authenticator is being used, it's the point where the application auth process switches from main org to sub org. At this point the flow is still being executed at the main org level application. But the url being generated here leads to sub org first factor page. If we add the spId in the context to the URL, it'll be the main org application id, which is not valid for the sub org application page.

PR #6083, opened 2024-10-31
Modify certificate validator to allow encoded PEM certificates

ashanthamara:
- This PR will allow to have encoded pem certificates to the certificate-mgt component. This is to facilitate the existing certificate management in the application-mgt component where the encoded certificate is stored in the IDN_CERTIFICATE table.
- Since we can directly convert the encoded the certificate string to the `X509Certificate` type without any additional operations this change will not add any additional overhead in the process.

### Related Issue:
- https://github.com/wso2/product-is/issues/21547


components/certificate-mgt/org.wso2.carbon.identity.certificate.management/src/test/resources/dbscripts/h2.sql
```diff
@@ -1,6 +1,6 @@
 CREATE TABLE IF NOT EXISTS IDN_CERTIFICATE (
             ID INTEGER NOT NULL AUTO_INCREMENT,
-            UUID CHAR(36) NOT NULL,
+            UUID CHAR(36),
```
malithie, line 3: why this not null is removed ?

ashanthamara: Removed not null to facilitate the migration requirements where we have to alter the table and add the new column. In such case it will not not work unless we give a default value for the UUID of the existing entries.
Hence we can either remove NOT NULL requirement from the column or have a default value for the existing entries.

ashanthamara: Since we are avoiding data migrations at this moment proceeding without having the NOT NULL for the UUID column is the most suitable option.

malithie: ack

malithie (approved):

ashanthamara:
Merging this without running integration tests since this component is not yet onboarded or used in the pack.

PR #6118, opened 2024-11-08
Fix exception when no actions configured.

malithie:
Resolves https://github.com/wso2/product-is/issues/21383

This PR fixes the case where an exception will thrown from ActionExecutionService when no actions are configured in the system.
So when no actions configured or if there's a failure in retrieving actions for the flow it will be considered as a scenario where no actions are applied. Thus, token issuance works as expected the state before.


ashanthamara (approved):

malithie:
> PR builder completed Link: https://github.com/wso2/product-is/actions/runs/11749025346 Status: **failure**

This failure is due to the fact that the version of the oauth component which is dependent on the framework components is not yet being updated in the product

PR #6164, opened 2024-12-01
Implement rule metadata service.

malithie:

Resolves https://github.com/wso2/product-is/issues/21803

In order to populate the rule UI it should be possible to retrieve all fields and field value options that are needed to configure expressions.
This PR implements the core rule metadata service.

Merging this PR releases the service component and the feature.
Feature will be bundled to the product over a separate PR. So note that merging this PR will not bundle the feature with the product.


malithie:
Looks like the pr builder is failing referring to an old commit

components/rule-mgt/org.wso2.carbon.identity.rule.metadata/src/main/java/org/wso2/carbon/identity/rule/metadata/config/RuleMetadataConfigFactory.java
```diff
@@ -0,0 +1,86 @@
...
+    public static FlowConfig getFlowConfig() {
+
+        return flowConfig;
+    }
+
+    private static File getFile(String fileName) throws RuleMetadataConfigException {
+
+        File file = new File(fileName);
+        if (!file.exists() || !file.isFile()) {
+            throw new RuleMetadataConfigException("File not found at: " + file.getAbsolutePath());
```
ashanthamara, line 81: Don't we need a fallback option when the file is missing ?
Eg: Reading a default file

malithie: There's nothing like default file. For things to function this data should be available and it's something product developers do

ashanthamara (approved):

PR #6171, opened 2024-12-03
Add versioning, created updated times for notification templates schema

darshanasbg:

$subject. Related to,
- https://github.com/wso2/product-is/issues/21887

can merge this pull request.]

### Follow up actions

Merge,
- TBD


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/dbscripts/postgresql.sql
```diff
@@ -1579,6 +1579,9 @@ CREATE TABLE IDN_NOTIFICATION_TYPE (
     NAME VARCHAR(255) NOT NULL,
     CHANNEL VARCHAR(255) NOT NULL,
     TENANT_ID INTEGER NOT NULL,
+    VERSION VARCHAR(15) NOT NULL,
+    CREATED_AT TIMESTAMP NOT NULL,
```
Thumimku, line 1583:
```suggestion
    CREATED_AT TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
    UPDATED_AT TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
```
Some tables support have DEFAULT and ON_UPDATE support, we can use that if applicable?

UdeshAthukorala (approved):

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/dbscripts/postgresql.sql
darshanasbg: Whether these values needs to be managed from the db side or the application side is something we have to decide here..

I also thought to handle this from the db level as per https://github.com/wso2/carbon-identity-framework/commit/ee52c3c35a6c745ed82641e66508e9e0827406c9

But bit further check on things, I choose maintaining these values at the application level due to following reasons: 
1. There as data format difference between databases.. But this is a minor reason, since the difference was only at the millisecond levels
2. Different db impls have their own way of doing this, there is no constant one way to do across this in different db types and versions.. and when consider an example on how these going to be  effected with the data replications with the things like DR or multi-regions is something we don't have in-depth expertise with the various database types we need to support...

So going down with the path of depending on db to maintain these records increase the complexity of the solution, reduce the maintainability product. So selected the handling at the application level data handling to make things simple.

If we are concerned about, each developer to having to handle this at each of their DAO impls, we can avoid duplicating such duplicated efforts by centralizing maintaining this logic in a place like [jdbcTemplates](https://github.com/wso2/carbon-utils/blob/master/components/org.wso2.carbon.database.utils/src/main/java/org/wso2/carbon/database/utils/jdbc/JdbcTemplate.java).

omindu: +1 for the approach suggested initially in this PR based on @darshanasbg reasoning

How were we historically handling the created times? The above concerns should be there for those as well

PR #6188, opened 2024-12-09
Introduce a config to allow system prefix for roles

ImalshaG:
With the current role management model, we do not allow the creation of roles with system_ prefix as it is used for the special roles created by the system in order to maintain the backward compatibility. But for IS KM connector only, we will be allowing the creation of roles with system_ prefix which will be allowed through the config:

```
[role_mgt]
allow_system_prefix_for_role = true
```
This config will be disabled by default for IS, and for IS KM it can be enabled if needed as above.


piraveena:
Can you link the product is issue here?

piraveena (approved):

PR #6225, opened 2024-12-19
Add a config to autofill the username with subject attribute while jit provisioning

DilshanSenarath:
## Purpose
This pr will introduce the changes require to configure the auto populating username with subject attribute.

Since the auto-population of the username field with the subject attribute may already be in use by other customers, we plan to introduce a new configuration to maintain backward compatibility while providing the requested change.

We’ll add a boolean server level configuration option called `AutofillUsernameFieldWithSubjectAttribute` within the JITProvisioning configuration section. Before appending the username to the registration UI redirection URL, we will first verify if the prompt for username, password, and consent provisioning is enabled. If enabled, we will then check if this new configuration is set to `true` and, if so, add the `username` parameter to the registration UI redirection URL.

```bash
[authentication.jit_provisioning]
autofill_username_field_with_subject_attribute = false
```

## Related Issue
- https://github.com/wso2/product-is/issues/21578


piraveena:
DO we need this config for master?

DilshanSenarath:
> DO we need this config for master?

Yes, this feature allows auto-populating the subject attribute value from the federated IDP during username and password provisioning. Therefore, we need to retain the existing behavior. However, a specific customer, who does not have a unique attribute other than the resource ID in their federated IDP, requested the ability to disable the auto-filling capability. As this could be a general requirement, the PR has been created to address it.

piraveena (approved):

PR #6269, opened 2025-01-07
Improve action management service to manage actions with rules.

malithie:

Resolves https://github.com/wso2/product-is/issues/22185

When it comes to extensions it should be possible to let to invoke extensions conditionally based on some rule applicable to the context.
This PR introduce the capability to manage such extensions (actions) with rules improving the action management service component to add/update actions with rules using rule management service.

Note audit logs for rule update in actions will be done in a future iteration: https://github.com/wso2/product-is/issues/22186


malithie:
> 2 New issues

These issues are not quite valid

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOFacade.java
```diff
@@ -294,6 +307,70 @@ private void addEncryptedAuthSecretsToBuilder(ActionDTOBuilder actionDTOBuilder,
...
+        } else if (existingActionDTO.getActionRule() != null && updatingActionDTOBuilder.getActionRule() !=
+                null) {  // This means the existing action rule is updated when updating the action.
+            try {
+                updatingActionDTOBuilder.getActionRule().getRule().setId(existingActionDTO.getActionRule().getId());
+                ActionMgtServiceComponentHolder.getInstance()
+                        .getRuleManagementService()
+                        .updateRule(updatingActionDTOBuilder.getActionRule().getRule(), tenantDomain);
+            } catch (RuleManagementException e) {
+                throw new ActionMgtServerException("Error while updating the Rule associated with the Action.", e);
+            }
```
ashanthamara, line 357: Shall we extract these to a separate method as well ?

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOImpl.java
```diff
@@ -415,6 +463,53 @@ private void updateAuthentication(String actionId, Authentication updatingAuthen
...
+                    Collections.singletonList(ActionMgtConstants.RULE_PROPERTY), tenantId);
+        } catch (TransactionException e) {
+            throw new ActionMgtServerException("Error while removing the reference for the Rule in Action.", e);
+        }
+    }
+
+    private ActionRule populateRule(Map<String, String> propertiesFromDB, Integer tenantId) {
+
+        if (!propertiesFromDB.containsKey(ActionMgtConstants.RULE_PROPERTY)) {
+            return null;
```
ashanthamara, line 506: Shall we add a comment saying that the rule is an optional attribute

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOFacade.java
```diff
@@ -72,10 +75,11 @@ public void addAction(ActionDTO actionDTO, Integer tenantId) throws ActionMgtExc
                 ActionDTOBuilder actionDTOBuilder = new ActionDTOBuilder(actionDTO);
                 // Encrypt authentication secrets
                 encryptAddingAuthSecrets(actionDTOBuilder);
+                // Add action rule
+                addActionRule(actionDTOBuilder, IdentityTenantUtil.getTenantDomain(tenantId));
```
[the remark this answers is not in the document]
malithie: Rule is constructed at the api layer using the utility the rule management service provides. So all validations are done at that level. Rule management service actually does not return any client errors. 
So there's no requirement to support for client errors here. That doesn't apply

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOFacade.java
```diff
@@ -294,6 +307,70 @@ private void addEncryptedAuthSecretsToBuilder(ActionDTOBuilder actionDTOBuilder,
...
+
+        if (actionDTOBuilder.getActionRule() != null) {
+            try {
+                ActionMgtServiceComponentHolder.getInstance()
+                        .getRuleManagementService()
+                        .addRule(actionDTOBuilder.getActionRule().getRule(), tenantDomain);
+            } catch (RuleManagementException e) {
+                throw new ActionMgtServerException("Error while adding the Rule associated with the Action.", e);
+            }
+        }
```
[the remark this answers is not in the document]
malithie: I don't see a real necessity of this block here as it is just one condition

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/util/ActionDTOBuilder.java
```diff
@@ -134,6 +138,17 @@ public EndpointConfig getEndpoint() {
         return this.endpoint;
     }
 
+    public ActionDTOBuilder rule(ActionRule rule) {
+
+        this.rule = rule;
+        return this;
+    }
+
+    public ActionRule getActionRule() {
```
ashanthamara, line 147:
```suggestion
    public ActionRule getRule() {
```
Do we need to repeat the Action in the method name again ?

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOFacade.java
```diff
@@ -294,6 +307,70 @@ private void addEncryptedAuthSecretsToBuilder(ActionDTOBuilder actionDTOBuilder,
...
+    }
+
+    private void updateActionRule(ActionDTOBuilder updatingActionDTOBuilder, ActionDTO existingActionDTO,
+                                  String tenantDomain)
+            throws ActionMgtException {
+
+        if (existingActionDTO.getActionRule() == null && updatingActionDTOBuilder.getActionRule() != null) {
+            // This means a new action rule is added when updating the action.
+            addActionRule(updatingActionDTOBuilder, tenantDomain);
+        } else if (existingActionDTO.getActionRule() != null && updatingActionDTOBuilder.getActionRule() == null) {
```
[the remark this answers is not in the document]
malithie: I didn't get you. This doesn't check other attributes. This just check for rule. So, rule doesn't need to be there in payload always

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOFacade.java
malithie: This method is already about rule update. so I don't think it's necessary

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOImpl.java
malithie: Is that necessary. The code says and the API says that

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/util/ActionDTOBuilder.java
malithie: yes. Because from ActionRule you can get the real Rule. So it should not be something like getRule().getRule()

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOFacade.java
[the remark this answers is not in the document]
ashanthamara: With current approach for updating action, some respective attributes in the payload can be null as they are not updating(Eg: When only the action name is updating, payload only contains the name; others are null).

Let's say we have a rule configured in the action, and we are updating the action name. In that case, rule will be null in the payload. Due to that this condition(`updatingActionDTOBuilder.getActionRule() == null`) will be true and it delete the rule.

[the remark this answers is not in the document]
malithie: ok so it's not a put and a patch. Then we have to think of a different way. Good catch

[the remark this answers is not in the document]
malithie: This is fixed as below.
- Current patching capability is preserved
- When a rule is not updated as the action updates, the payload should not include the rule object.
- When rule is added or updated, it should send the rule object with it's expressions and other rules
- When rule is removed, the payload should include an empty rule object. e.g., `rule:{}`

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOImpl.java
```diff
@@ -106,8 +115,12 @@ public ActionDTO getActionByActionId(String actionType, String actionId, Integer
         if (actionBuilder == null) {
             return null;
         }
-        actionBuilder.setEndpointAndProperties(getActionPropertiesFromDB(actionId, tenantId));
```
ashanthamara, line 109: As we don't use `setEndpointAndProperties()` anymore, let's remove that from the ActionDTOBuilder class

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOFacade.java
```diff
@@ -294,6 +307,108 @@ private void addEncryptedAuthSecretsToBuilder(ActionDTOBuilder actionDTOBuilder,
                         .build());
     }
 
+    private void addActionRule(ActionDTOBuilder actionDTOBuilder, String tenantDomain) throws ActionMgtException {
+
+        if (actionDTOBuilder.getActionRule() == null || actionDTOBuilder.getActionRule().getRule() == null) {
```
ashanthamara, line 312: Shall we remove these if condition as these are checked in the `isAddingNewActionRule()` method ?

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOFacade.java
```diff
@@ -294,6 +307,108 @@ private void addEncryptedAuthSecretsToBuilder(ActionDTOBuilder actionDTOBuilder,
...
+                    .getRuleManagementService()
+                    .updateRule(updatingActionDTOBuilder.getActionRule().getRule(), tenantDomain);
+        } catch (RuleManagementException e) {
+            throw new ActionMgtServerException("Error while updating the Rule associated with the Action.", e);
+        }
+    }
+
+    private void deleteActionRule(ActionDTO actionDTO, String tenantDomain) throws ActionMgtServerException {
+
+        if (actionDTO.getActionRule() == null) {
```
ashanthamara, line 399: Same as above comment

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOFacade.java
malithie: Well this is used in addAction not just update. So we can't remove that from here. Taking this out also impacts readability. So let's keep it.

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOFacade.java
malithie: Same reply applies

components/action-mgt/org.wso2.carbon.identity.action.management/src/main/java/org/wso2/carbon/identity/action/management/dao/impl/ActionManagementDAOImpl.java
malithie: Fixed https://github.com/wso2/carbon-identity-framework/pull/6269/commits/dbc0b1d635d2ee72742509dee496d7a8ba755a39

ashanthamara (approved):

malithie:
> PR builder completed Link: https://github.com/wso2/product-is/actions/runs/12682779106 Status: **failure**

Failing test case here is actually passing locally

malithie:
Merged as the failing test case is passing locally supposing it's an intermittent failure.

PR #6295, opened 2025-01-16
Provide a configuration to skip internal prefix append to the roles claim of the user info response

sadilchamishka:
The oauth2/userinfo response provides roles claim with an inconsistent manner.

1. If one role exist, it will be as   "roles" : "role1"
2. If multiple role exist, it will be as.    "roles" : ["Internal/role1", "Internal/role2"]

Having the internal prefix of the userinfo response can cause issues in some cases. As the internal prefix is not exposed to outside, this prefix should be removed. 

But there can be customer adaption for this behaviour. Hence the correct behaviour as the default behaviour of the identity server should be provided during major version. 

To enable this improvement, add the following configuration to the deployment.toml file
```
[oauth.oidc.user_info]
remove_internal_prefix_from_roles=true
```

### Related Issues
- https://github.com/wso2/product-is/issues/22088

### DIFF
--- a/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml
+++ b/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml
@@ -586,6 +586,7 @@
             <UserInfoEndpointClaimRetriever>org.wso2.carbon.identity.oauth.endpoint.user.impl.UserInfoUserStoreClaimRetriever</UserInfoEndpointClaimRetriever>
             <UserInfoEndpointRequestValidator>org.wso2.carbon.identity.oauth.endpoint.user.impl.UserInforRequestDefaultValidator</UserInfoEndpointRequestValidator>
             <UserInfoEndpointAccessTokenValidator>org.wso2.carbon.identity.oauth.endpoint.user.impl.UserInfoISAccessTokenValidator</UserInfoEndpointAccessTokenValidator>
+            <UserInfoInternalPrefixedRolesClaimAllowed>true</UserInfoInternalPrefixedRolesClaimAllowed>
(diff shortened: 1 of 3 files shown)


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml
```diff
@@ -586,6 +586,7 @@
             <UserInfoEndpointClaimRetriever>org.wso2.carbon.identity.oauth.endpoint.user.impl.UserInfoUserStoreClaimRetriever</UserInfoEndpointClaimRetriever>
             <UserInfoEndpointRequestValidator>org.wso2.carbon.identity.oauth.endpoint.user.impl.UserInforRequestDefaultValidator</UserInfoEndpointRequestValidator>
             <UserInfoEndpointAccessTokenValidator>org.wso2.carbon.identity.oauth.endpoint.user.impl.UserInfoISAccessTokenValidator</UserInfoEndpointAccessTokenValidator>
+            <UserInfoInternalPrefixedRolesClaimAllowed>true</UserInfoInternalPrefixedRolesClaimAllowed>
```
PasinduYeshan, line 589: What about using a name like <UserInfoRemoveInternalPrefixFromRoles>? It seems clearer and directly describes the action of removing the "Internal" prefix from roles. 
The current name, UserInfoInternalPrefixedRolesClaimAllowed, feels more like it’s about not allowing internal prefixed roles at all, which might be misleading. WDYT?

sadilchamishka: Fixed by - https://github.com/wso2/carbon-identity-framework/pull/6295/commits/69042ffdabe3a91ad689e45b95b9d62f24916382

PasinduYeshan (approved):

PR #6296, opened 2025-01-16
Improve authenticator mgt to store image and description.

Thisara-Welmilla:
Issue:
- https://github.com/wso2/product-is/issues/22273

Currently there is no way to configure image_url and description of the user defined local authenticators. The federated authenticators, those can be configured for the assocaited identity providers. This PR is to provide support for that for user defined local authenticators.


components/application-mgt/org.wso2.carbon.identity.application.common/src/main/java/org/wso2/carbon/identity/application/common/ApplicationAuthenticatorService.java
```diff
@@ -36,9 +36,9 @@
 import java.util.ArrayList;
 import java.util.List;
 
+import static org.wso2.carbon.identity.application.common.constant.AuthenticatorMgtSQLConstants.Column.DISPLAY_NAME;
```
malithie, line 39: We should not import the column name at the service level. what added this ?

malithie (approved):

PR #6334, opened 2025-01-22
Improve framework to skip retry from authenticator

ZiyamSanthosh:
Currently, if an authenticator fails, the reason for the failure will be shown when the authenticator is re-executed. For this, the authenticator depends on the retryAuthenticationEnabled function. The resend mechanism also depends on the same method. For scenarios where the user denies the authentication flow from the authenticator, the user has to be redirected to the application with an error message without retrying the authenticator. 

This improvement addresses the above mentioned concern.


ZiyamSanthosh:
To cover the newly added lines, AbstractAppAuthSkipRetryTest.java was introduced. Even though the tests covers the new lines, Codecov is not detecting it. I have locally checked that the test are covering the new lines added.

ZiyamSanthosh:
The last commit adds a comment to the file. The previous PR builder passed with the newly added changes.
Screenshot 2025-01-24 at 19 10 30

janakamarasena (approved):

PR #6419, opened 2025-02-03
Restrict shared apps to only use shared roles when shared apps are accessed

ShanChathusanda93:

- $subject

### DIFF
--- a/components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/listener/DefaultRoleManagementListener.java
+++ b/components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/listener/DefaultRoleManagementListener.java
@@ -46,7 +46,6 @@ import org.wso2.carbon.identity.role.v2.mgt.core.model.IdpGroup;
 import org.wso2.carbon.identity.role.v2.mgt.core.model.Permission;
 import org.wso2.carbon.identity.role.v2.mgt.core.model.Role;
 import org.wso2.carbon.identity.role.v2.mgt.core.model.RoleBasicInfo;
-import org.wso2.carbon.identity.role.v2.mgt.core.model.RoleProperty;
 import org.wso2.carbon.identity.role.v2.mgt.core.model.UserBasicInfo;
 
 import java.util.ArrayList;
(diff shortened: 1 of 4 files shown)


components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/listener/DefaultRoleManagementListener.java
```diff
@@ -459,6 +465,45 @@ public void preGetRoleIdListOfGroups(List<String> groupIds, String tenantDomain)
...
+            return;
+        }
+
+        filterAndRemoveNonSharedRoles(roleIds, tenantDomain);
+    }
+
+    private static boolean isFragmentApplication() {
+
+        boolean isFragmentApp = false;
+        if (IdentityUtil.threadLocalProperties.get().get(ApplicationConstants.IS_FRAGMENT_APP) != null) {
```
AnuradhaSK, line 478: From where this threadlocal set and unset

components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/listener/DefaultRoleManagementListener.java
```diff
@@ -447,6 +448,11 @@ public void preGetRoleIdListOfUser(String userId, String tenantDomain) throws Id
     public void postGetRoleIdListOfUser(List<String> roleIds, String userId, String tenantDomain)
             throws IdentityRoleManagementException {
 
+        if (!isFragmentApplication()) {
+            return;
+        }
+
+        filterAndRemoveNonSharedRoles(roleIds, tenantDomain);
```
SujanSanjula96, line 455: I think it will be better to implement this logic in `getAssociatedRolesOfApplication` method in `ApplicationManagementService` which is used by in `getRolesAssociatedWithApplication ` method in `AppAssociatedRolesResolverImpl.java`. Relying on thread local properties in service level will not be an ideal approach since we can't guarantee setting thread local properties in every place that this service is used.

ShanChathusanda93:
Closing the PR since this will be handled from [1]

[1] https://github.com/wso2-extensions/identity-inbound-auth-oauth/pull/2703

PR #6433, opened 2025-02-05
Add indexes for IDP_AUTHENTICATOR in SQL scripts

Sadeesha-Sath:

- Add Indexes to IDP_ID and DEFINED_BY Columns of IDP_AUTHENTICATOR Table to fix performance issues

### Follow up actions

- Run a Perf test


Sadeesha-Sath:
- IDP_ID is a foreign key of the Table IDP_AUTHENTICATOR.
 - However some databases like MSSQL, PostgreSQL, Oracle does not create an implicit index for foreign keys.
 - Databases that create implicit indexes for foreign keys like MySQL also overwrites the implicit index when creating an explicit index with the same column.

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/dbscripts/db2.sql
```diff
@@ -2541,6 +2541,12 @@ CREATE INDEX IDN_SECRET_TYPE_ID ON IDN_SECRET (TYPE_ID);
 CREATE INDEX IDX_CLAIM_TI_CU ON IDN_CLAIM (TENANT_ID, CLAIM_URI)
 /
 
+-- IDP_AUTHENTICATOR --
+CREATE INDEX IDX_AUTH_DEFINED_BY ON IDP_AUTHENTICATOR (DEFINED_BY)
```
ashanthamara, line 2545:
```suggestion
CREATE INDEX IDX_AUTH_DEFINED_BY ON IDP_AUTHENTICATOR (DEFINED_BY)
```
Shall we rename, IDX_IDP_AUTH_DEFINED_BY 
as there are few tables having AUTH in there name. (Eg: SP_AUTH_SCRIPT, SP_AUTH_STEP etc)

Sadeesha-Sath: Renamed

ashanthamara (approved):

PR #6451, opened 2025-02-07
Exclude adding the spid to external calls

janakamarasena:
Earlier the spId was added to external calls. This PR excludes adding the spid to external calls. 
The code specifically handling the Organization authenticator is removed as it is a federated authenticator it is already covered by the isExternalCall check.

fixes: https://github.com/wso2/product-is/issues/22667


janakamarasena:
Integration tests passed successfully https://github.com/janakamarasena/product-is/actions/runs/13196618626/job/36839257365

PR #6492, opened 2025-02-14
 Check claim uniqueness without duplicate claim check

sahandilshan:
### Purpose
- Do the `Password can't be equal to login attributes` policy check correctly by only validating if the password is equal to a unique claim. It won't check if the unique claim is duplicated with other user
- As a fix to https://github.com/wso2/product-is/issues/12699, we have sent the following https://github.com/wso2/carbon-identity-framework/pull/3812, where it calls the `checkClaimUniqueness` method to check if the password is equal to a login attribute. The problem here is that the `checkClaimUniqueness` is doing additional validation where it checks if the login attributes have been duplicated in the same userstore. So if two users share the same login attribute values (ex: same emails), when the password is getting updated, it will first check
    1. If the password is equal to a login attribute (if the user hasn't used a login attribute as the password this logic will be passed)
    2. Then it checks if the login attribute is duplicated in the same userstore (For a password reset scenario, this is not needed). password reset gets an error due to this logic. 

In this PR, we have introduced a new method that only does the first check and ignores the second check for admin-initiated password reset scenarios 

### Related Issue(s)
- https://github.com/wso2/product-is/issues/22677


piraveena (approved):

piraveena:
Please add unit tests

piraveena (approved):

PR #6503, opened 2025-02-15
Add profile update flow to Identity Context and update the password action builders

ashanthamara:

1. This will add new PROFILE_UPDATE flow to the Identity Context flows
2. Update the password update request builder and the rule evaluation data provider to support those flows

### Related Issue
-


ashanthamara:
Changes are related to adding new flow type to the Identity context. Hence can proceed with the code cov failure

malithie (approved):

PR #6506, opened 2025-02-15
Handle session retrieval issue for sub org users

SujanSanjula96:
This PR is introduced to handle session retrieval issues for sub org users. Currently only the session created in sub org is shown when getting the session list of sub org list. Ideally session created in primary org should be listed and sub org session should not be listed.

When implementing the fix, following cases are considered.
1. Root org normal apps - In this case, session is created for the root org and it should be considered as an effective session since the session retrieval request is also coming from the same root org.
2. B2B applications -  In this case, sessions are created in both root org and B2B org and root org session should be considered as the effective session even though session retrieval request comes from sub org.
3. SaaS applications - In this case, if a user logs into a SaaS app in different tenant, session is created in that tenant. But session retrieval request comes from user resident tenant. But we should return this session as an effective session.
4. Same userstore plugged in different tenants - In this case, same user ID can have different sessions in different tenants. Session retrieval requests can come from different tenants. Here, the sessions belong to the tenant related to session retrieval request should be considered as effective sessions.

Before the fix [1], the behaviour was,
- 1 & 3 cases working as expected.
- Both sessions were listed in case 2.
- All the sessions in different tenants were shown in case 4.

After the fix [1] which is done to address the case 4,
- 1 & 4 cases were handled correctly. 
- Only sub org session is shown in case 2.
- Only the sessions related to session request tenant domain are shown in case 3.

With this fix, we handled the above four cases to have expected behaviour.

[1] - https://github.com/wso2/carbon-identity-framework/pull/5693

Related Issue - https://github.com/wso2/product-is/issues/22310

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
@@ -626,13 +626,38 @@ public class UserSessionManagementServiceImpl implements UserSessionManagementSe
             String loginTenantDomain = FrameworkUtils.getLoginTenantDomainFromContext();
             boolean isOrganization = OrganizationManagementUtil.isOrganization(loginTenantDomain);
             if (StringUtils.equals(sessionTenantDomain, loginTenantDomain)) {
-                return !(isOrganization & areAllFragmentAppsInUserSession(userSession));
-            } else {
-                if (isOrganization &&
-                        validatePrimaryOrganization(sessionTenantDomain, loginTenantDomain)) {
+                // This block handles the scenario where session tenant domain and login tenant domain are equal.
+                if (isOrganization & areAllFragmentAppsInUserSession(userSession)) {
+                   /*
+                   When a sub org user authenticates using a shared application, sessions are created for shared app in
+                   sub organization and parent app in primary organization. In this case, the session created in primary
+                   organization is the effective session. Hence, this session is ignored for the session list.
+                    */
+                    return false;
+                } else {
                     return true;
                 }
-                return isSaaSAppAvailableInUserSession(userSession);
+            } else {
+                // This block handles the scenario where session tenant domain and login tenant domain are different.
+                if (isOrganization &&
+                        validatePrimaryOrganization(sessionTenantDomain, loginTenantDomain)) {
+                    /*
+                    When a sub org user authenticates using a shared application, sessions are created for shared app in
+                    sub organization and parent app in primary organization. In this case, the session created in
+                    primary organization is the effective session. Hence, this scenario is considered as an
+                    effective session.
+                     */
+                    return true;
+                }
+                if (isSaaSAppAvailableInUserSession(userSession)) {
+                    /*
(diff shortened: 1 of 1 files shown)


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
```diff
@@ -605,10 +609,86 @@ private List<UserSession> getActiveSessionList(List<String> sessionIdList, Strin
...
+                    .anyMatch(property -> IS_FRAGMENT_APP.equals(property.getName()) &&
+                    Boolean.parseBoolean(property.getValue()));
+            if (!isFragmentApp) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    private boolean validatePrimaryOrganization(String sessionTenantDomain, String loginTenantDomain)
```
AnuradhaSK, line 664:
```suggestion
    private boolean isSessionTenantPrimaryOrgForLoginTenant(String sessionTenantDomain, String loginTenantDomain)
```

SujanSanjula96: Fixed with https://github.com/wso2/carbon-identity-framework/pull/6506/commits/f324397816b07d1799ff17ce79e7a7f6f60e67fe

AnuradhaSK (approved):

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
```diff
@@ -605,10 +609,111 @@ private List<UserSession> getActiveSessionList(List<String> sessionIdList, Strin
...
+            if (StringUtils.equals(sessionTenantDomain, loginTenantDomain)) {
+                // This block handles the scenario where session tenant domain and login tenant domain are equal.
+                if (isOrganization & areAllFragmentAppsInUserSession(userSession)) {
+                   /*
+                   When a sub org user authenticates using a shared application, sessions are created for shared app in
+                   sub organization and parent app in primary organization. In this case, the session created in primary
+                   organization is the effective session. Hence, this session is ignored for the session list.
+                    */
+                    return false;
+                } else {
```
pamodaaw, line 637: Do we need to have the else block here?

pamodaaw (approved):

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
SujanSanjula96: We can get rid of this else block and return true after all if else blocks. But kept it this way to separate every case clearly.

SujanSanjula96:
Merging since test failure is an intermittent test failure reported in https://github.com/wso2/product-is/issues/23074

PR #6546, opened 2025-02-21
Add new config for default userStore for authentication.

Thisara-Welmilla:
Issue:
- https://github.com/wso2/product-is/issues/23211

Add the following new configuration to set the default userstore for local users authenticating via the external service when the userstore domain is not specified in the response from the external authenticator.
```
Actions.Types.Authentication.DefaultUserStore
```

### DIFF
--- a/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
+++ b/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
@@ -2145,6 +2145,7 @@
             </PreIssueAccessToken>
             <Authentication>
                 <Enable>{{actions.types.authentication.enable}}</Enable>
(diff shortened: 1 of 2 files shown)


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -2145,6 +2145,7 @@
             </PreIssueAccessToken>
             <Authentication>
                 <Enable>{{actions.types.authentication.enable}}</Enable>
+                <DefaultUserStore>{{actions.types.authentication.default.userStore}}</DefaultUserStore>
```
malithie, line 2148: Do we want to break this as default.userStore ?
We won't have anything under default further

why not go with defaultUserStore as one key word

malithie (approved):

PR #6562, opened 2025-02-25
Update userId claim in the custom federated authenticator template to use with JIT provisioning

Shenali-SJ:
### Purpose
This PR updates the userId claim of the custom federated authenticator connection template to default username claim URI. This will be utilized in JIT provisioning.


Thisara-Welmilla (approved):

Shenali-SJ:
Closing this as this requirement is only valid for the cloud.

PR #6653, opened 2025-03-27
Changing the SharedProfileValueResolvingMethod for lastPasswordUpdateTime Claim

BimsaraBodaragama:
### Purpose
Change the `SharedProfileValueResolvingMethod` from **FromOrigin** to **FromSharedProfile** to resolve the value using shared profile data.

### Description
When password expiry is enabled, inviting or sharing a user results in the addition of a shared user. During this process, the system attempts to update the `lastPasswordUpdateTime` claim for the shared user. However, since `lastPasswordUpdateTime` is an identity claim, it is generally configured with the `FromOrigin` resolving method, preventing updates from the shared profile. As a result, this causes an error when trying to update the claim for a shared user.  

To address this issue, the `SharedProfileValueResolvingMethod` for the `lastPasswordUpdateTime` claim needs to be changed from `FromOrigin` to `FromSharedProfile`. This allows the claim to be updated using data from the shared profile, eliminating the error.  

Since shared users are not used for direct application logins, a dummy password is generated at the time of user creation due to the mandatory nature of the password field. Consequently, the password expiry update does not impact the core functionality or introduce any risk. Unlike other identity claims that remain specific to the user, the password-related information is consistent with the shared profile. After a thorough discussion, it was concluded that resolving the `lastPasswordUpdateTime` claim using the shared profile is a suitable solution.  

### Goals
- Update the `lastPasswordUpdateTime` claim to resolve using shared profile data.
- Prevent errors when updating the claim for shared users.
- Ensure seamless functionality when password expiry is enabled.
- Maintain alignment with the system's shared user management approach.

### Approach
- Update `SharedProfileValueResolvingMethod` from `FromOrigin` to `FromSharedProfile`.  
- Validate the change and ensure no other claims are affected.  
- Perform tests to confirm correct claim resolution and error prevention.

### Related Issues
product-is issue: https://github.com/wso2/product-is/issues/23604
- The above issue is a front port of the effort originally tracked for IS-7.0.0 under https://github.com/wso2/product-is/issues/23441.  
- Separate issues have been created to address the differences in behavior between IS-7.0.0 and IS-7.1.0 onwards.
    - https://github.com/wso2/product-is/issues/23604
    - https://github.com/wso2/product-is/issues/23603

### DIFF
--- a/features/claim-mgt/org.wso2.carbon.claim.mgt.server.feature/resources/conf/claim-config.xml
+++ b/features/claim-mgt/org.wso2.carbon.claim.mgt.server.feature/resources/conf/claim-config.xml
@@ -731,7 +731,7 @@
 				in your user store must be configured for this. -->
 				<AttributeID>lastPasswordUpdate</AttributeID>
 				<Description>Last Password Update Time</Description>
-				<SharedProfileValueResolvingMethod>FromOrigin</SharedProfileValueResolvingMethod>
+				<SharedProfileValueResolvingMethod>FromSharedProfile</SharedProfileValueResolvingMethod>
 			</Claim>
 			<Claim>
 				<ClaimURI>http://wso2.org/claims/displayName</ClaimURI>


features/claim-mgt/org.wso2.carbon.claim.mgt.server.feature/resources/conf/claim-config.xml
```diff
@@ -731,7 +731,7 @@
 				in your user store must be configured for this. -->
 				<AttributeID>lastPasswordUpdate</AttributeID>
 				<Description>Last Password Update Time</Description>
-				<SharedProfileValueResolvingMethod>FromOrigin</SharedProfileValueResolvingMethod>
+				<SharedProfileValueResolvingMethod>FromSharedProfile</SharedProfileValueResolvingMethod>
```
sadilchamishka, line 734: When other identity claims are managed from the origin, why we need to manage this identity claim only from the shared profile?
Won't this would be a future inconsistency?

For an example, you can't update any identity claims of shared users except this claim. When shared user password is generated one and not meant to login even, this change might cause questions.

We may have to block password reset for the shared users, its looks we can. If we are blocking, identity is always managed from the parent hence this treatment looks incorrect.

BimsaraBodaragama: Hi @sadilchamishka,

Identity claims are specialized attributes that convey critical information about a user's account status or state, such as `lastLoginTime`, `accountDisabled`, and `accountLocked`. As you said, these claims are typically managed at the originating organization level, ensuring centralized control over essential user account attributes.

However, the introduction of shared user profiles introduces unique considerations. Shared user profiles are essentially shadow accounts that enable users from a parent organization to access resources in child organizations. These profiles are not intended for direct authentication. Instead, they are used to switch between organizations. Despite this, the system mandates the creation of a password for these shared profiles, leading to the assignment of dummy passwords. 

This requirement, while procedural, can inadvertently trigger issues when functionalities like password expiry policies are enforced. Unlike typical scenarios, password expiry can be selectively applied to child organizations [[1]](https://github.com/wso2/product-is/issues/23441#:~:text=Enable%20password%20expiry%20at%20the%20sub%2Dorganization%20level%20using%20the%20Identity%20Governance%20API%3A), even if it is not applied at the parent organization level. As a result, when a password expiry policy is enabled for a child organization, the system attempts to update the `lastPasswordUpdateTime` claim for the shared user profiles. Given that identity claims are traditionally resolved from the origin, this update process encounters conflicts, resulting in errors during user invitation and sharing workflows [[2]](https://github.com/wso2/product-is/issues/23441).

Addressing this challenge at a governance level, such as conditionally bypassing updates for shared users when password expiry is active, proves to be impractical. Different enterprises may have implemented custom connectors with bespoke properties, making it cumbersome to account for all variations at a higher level. Therefore, implementing a solution at the core level ensures a more robust and universally applicable fix.

Furthermore, the role of shared user profiles is evolving. Future enhancements may necessitate the ability to lock specific shared user accounts, implying that some identity claims like `accountLocked` should be managed within the shared profile context rather than at the origin. This progression underscores the need to reconsider the blanket approach of managing all identity claims solely at the parent organization level.

In light of these developments, transitioning the resolving method of the `lastPasswordUpdateTime` claim from `FromOrigin` to `FromSharedProfile` is a strategic and necessary adjustment. This change not only addresses current conflicts but also aligns with anticipated requirements, ensuring that WSO2 IS remains adaptable and responsive to the needs of complex organizational structures.

[1] [Enable password expiry at the sub-organization level using the Identity Governance API](https://github.com/wso2/product-is/issues/23441#:~:text=Enable%20password%20expiry%20at%20the%20sub%2Dorganization%20level%20using%20the%20Identity%20Governance%20API%3A)
[2] https://github.com/wso2/product-is/issues/23441

AnuradhaSK (approved):

features/claim-mgt/org.wso2.carbon.claim.mgt.server.feature/resources/conf/claim-config.xml
sadilchamishka: I am -1 for this and propose to get approve from @ashensw

AnuradhaSK: @sadilchamishka,
For shared users, we can't always say the identity claims are resolved from origin.

For eg: Having `FromOrigin` for Last logon time also questionable now. shared users can't login through username password option directly to sub org, but this shared user can logged in to sub org direactly via password options, like linking to federated login, SMS or Email OTP as first factor. This is an expected behavior too. If the shared user login to sub org directly through this option and  Last logon time updates on that time correctly, that claim also should be changed as resolving from shared profile.

When it comes to last password update time, as a claim it is not relavant to shared user and usually it should be ignored. However, due to the limitation of the product on unable to create the user without a password we have to go along with dummy passwords. This is the same approach as JIT provisioning flow. @BimsaraBodaragama Do we update JIT provisioning user's password change time or skip it conditionally in the password expiry related governance connector?

AnuradhaSK: Regarding password update time: Even though there is an architecture issue on unable to create a use without a password, we want to treat this shared user's original password is always from origin. So, let's keep this last password update time to be resolved from "FromOrigin" and handle the case from code level to avoid confusions on password expired user retrieval etc (Ideally shared users should be ignored when getting password expiring users, but this is something client needs to handle)

sadilchamishka: Let's assume there is a use case to lock the shared user hence he can't switch to sub-org space until unlock. We have to improve switching logic and having identity claim managed from org looks make sense. But making it default as `FromSharedProfile` or guide them to update it as per the use case is the concern I have.

The similar concern is raised for the `FromOrigin` claim, which I didn't identify the flow correctly yet. Checking whether we should make such claims `FromSharedProfile` by default or let admins to update it based on the use case they have.

Seeing default set to `FromSharedProfile` for a claim which is used for b2c cases might confuse about its meaning.

BimsaraBodaragama:
**Following further discussions based on https://github.com/wso2/carbon-identity-framework/pull/6653#discussion_r2020420837, this PR will be closed, and the effort will be tracked in the following PR:**  
https://github.com/wso2-extensions/identity-organization-management/pull/499 

cc: @AnuradhaSK, @SujanSanjula96, @sadilchamishka, @ashensw

PR #6657, opened 2025-04-01
Add API resource access to notification sender v2 API

lashinijay:

> $subject

### Related Issue

- https://github.com/wso2/product-is/issues/23405

### DIFF
--- a/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/resource-access-control-v2.xml.j2
+++ b/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/resource-access-control-v2.xml.j2
@@ -119,59 +119,31 @@
             <Scopes>internal_config_mgt_delete</Scopes>
         </Resource>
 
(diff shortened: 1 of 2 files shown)


features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/resource-access-control-v2.xml.j2
ThaminduR, line 1: rather than introducing new entries we could update the existing one to handle both v1 and v2

ThaminduR:
Update the xml file as well

ThaminduR (dismissed):
These changes will enable the v2 API endpoints while continuing to use the existing `internal_notification_senders_` scopes.

However, one thing to note is that the API resource representation will still only reflect the v1 version. This means that anyone intending to consume the v2 API will still need to authorize the v1 API in their application in order to obtain the necessary scopes.

ThaminduR (approved):

PR #6658, opened 2025-04-01
Add field to add custom properties for session context

mpmadhavig:
Related Issue:
- https://github.com/wso2/product-is/issues/23462


mpmadhavig:
This can be done using the sessionContext properties.

ThaminduR:
> This can be done using the sessionContext properties.

Shall we introduce this as a first class property @mpmadhavig ?

PR #6672, opened 2025-04-09
Add proper status for failed nodes.

inthirakumaaran:

Fix flow status in the fail() function


inthirakumaaran:
@PasinduYeshan 

However, in the standard (UI-based) login flow, we maintain the current behavior (status code will be `200`). For app-native authentication, this would be a significant change but one we should accept in order to ensure the correct behavior. On that note, we don't need a `flowStatus` for the `FAIL_COMPLETED` flow.

Lakshan-Banneheke:
Since this change only impacts users on version 7.0.0 and above, specifically when using the fail() function in adaptive scripts with app-native authentication, it has been agreed upon by the stakeholders to make a configuration available on demand and proceed with the proposed fix without introducing a config.

pamodaaw (approved):

PR #6770, opened 2025-05-21
Add missing WorkflowErrorConstants class

sadilchamishka:

$subject

### Related Issues
- https://github.com/wso2/product-is/issues/23374


sadilchamishka:
The `WorkflowErrorConstants` was an existing class in the framework repo. The duplicates are detected due to there exist error classes with the same structure. 
In order to avoid code duplications, we should be able to bring the ErrorConstants logic to a common module. Will address it in a separate effort.

Also the integration test runner is not required as this is a new class addition.

PR #6800, opened 2025-05-29
DataType attribute metadata related improvements

sadilchamishka:

This PR contains following improvements.

1. Some of the features we have given via additional properties of the attributes have been bring as first class properties.
ex: Uniqueness, multivalued, dataType, subAttributes, canonicalValues
Previous claim API response included those values in the additional properties of the attributes response (If configured).
With the recent improvements, the same information is received as first class meta data instead of additional properties. 
A configuration is given to provide the to preserve the previous behaviour.

```toml
[attribute.return_previous_additional_properties]
enable=true
```

2. The data type meta data of the system claims should be updated as per the improvement of data type support.
The default data type is "string" hence it is not explicitly mentioned in the claim-config.xml file. All other data types are defined against the each claims. The data type will be helpful for the presentation layer to visualize the attributes without having hard coded logics for each attributes.

3. The attributes are not stored in the database, but if the updates of the attributes are persisted. When we introduce new metadata for claims, the current algorithm of merging the claim properties from file and db has an issue of not reflecting the new meta data introduced from the claim-config.xml file for the system claims.

### Related Issues
- https://github.com/wso2/product-is/issues/24049


sadilchamishka:
The failing test case is not related to the changes in the PR. The test  failures are reported for the previous PRs also.

Screenshot 2025-05-31 at 19 14 14

PR #7022, opened 2025-07-25
Handle the accepted status code in the API client.

Malith-19:
### Purpose
- $subject to catch the workflow pending scenario when API clients invoked.

### Related issues
- https://github.com/wso2/product-is/issues/23374


Malith-19:
Test runner is pending in the fork: https://github.com/Malith-19/product-is/actions/runs/16524985569

PR #7210, opened 2025-08-21
Preserve backward comaptibility for pre issue access token action in allowed headers and params

ashanthamara:

> $subject

### Related Issue
-


components/action-mgt/org.wso2.carbon.identity.action.execution/src/main/java/org/wso2/carbon/identity/action/execution/internal/util/RequestFilter.java
```diff
@@ -67,6 +67,11 @@ public static List<Header> getFilteredHeaders(List<Header> requestHeaders, List<
             allowedHeaders.addAll(allowedHeadersInAction);
         } else if (hasServerAllowedHeaders) {
             allowedHeaders.addAll(allowedHeadersInServer);
+        } else if (ActionType.PRE_ISSUE_ACCESS_TOKEN.equals(actionType)) {
```
DMHP, line 70: Do we need to preserve the backward compatibility only for pre-issue_access token? What about the pre_profile_update as it was also available in older versions?

indeewari (approved):

PR #7254, opened 2025-08-25
[NEXT] Mask only the scope permission changes when scim v3 config is added

mpmadhavig:
With this PR below changes will be done.

> console new permission `console:role_assignments` & `console:org:role_assignments` permissions will be available in the system. 
    - They will not be useful as the permission changes are not reflected. (Need to enable by the config)
    - This is required for the migration.

Things depend on the config:
> Roles V3 API will be available. 
> After this change, by enabling config, the console roles created with `console:roles` & `console:org:roles` will loose the assignment permission if not specifically added via migration.

The above actions are performed before as they are not affecting the backward compatibility.

Related Issues:
- https://github.com/wso2/product-is/issues/21686

Related PRs:
- https://github.com/wso2/carbon-identity-framework/pull/7253


mpmadhavig:
https://github.com/wso2/carbon-identity-framework/pull/7260

PR #7480, opened 2025-10-03
Fix account linking logic to ensure secondary property is validated c…

ImalshaD:
**Related Issues**

- https://github.com/wso2/product-is/issues/25912

This pull request makes a small logic change in the `populateAccountLookupAttributes` method to improve the handling of secondary account lookup attribute mappings.

* Logic updated to only add the secondary account lookup attribute mapping if the property is not null and its value is not blank, instead of adding it when the property is null or the value is not blank.


ImalshaD:
Integration Tests run at the fork

- https://github.com/ImalshaD/product-is/actions/runs/18226082816
[Integration Test Runner](https://github.com/ImalshaD/product-is/actions/workflows/integration-test-runner-JDK11.yml)

ThaminduR (approved):

PR #7483, opened 2025-10-05
Support action versioning for action execution

Thisara-Welmilla:
Issue:
- https://github.com/wso2/product-is/issues/25738

With this PR:
- Introducing ActionVersioningHandler interface to handle the action version related logic. A default implementation DefaultActionVersioningHandler is provide for the action execution core. If any downstream component required custom logic, DefaultActionVersioningHandler can be extend and override with the custom logic.
- During the action execution, before trigger the action, it will check whether action version is retired or not eligible to trigger based on flow context. If not, action execution will be skipped.


components/action-mgt/org.wso2.carbon.identity.action.execution/src/main/java/org/wso2/carbon/identity/action/execution/api/service/ActionVersioningHandler.java
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Check whether the action version is eligible to be triggered for the action version and the given flow context.
+     *
+     * @param actionType  ActionType.
+     * @param action      Action.
+     * @param flowContext FlowContext.
+     * @return true if the action version is eligible to be triggered, false otherwise.
+     * @throws ActionExecutionException ActionExecutionException.
+     */
+    boolean isVersionEligibleForTrigger(ActionType actionType, Action action, FlowContext flowContext)
```
ashanthamara: How about having,
`canHandle()` or `canExecute()`

ashanthamara: I think it's better to send the `ActionExecutionRequestContext` without sending the Action.

And as we are registering `ActionVersioningHandler` classes against the action type, do we really need to pass action type to the method again ?

components/action-mgt/org.wso2.carbon.identity.action.execution/src/main/java/org/wso2/carbon/identity/action/execution/internal/service/impl/ActionExecutorServiceImpl.java
```diff
@@ -296,6 +306,30 @@ private boolean evaluateActionRule(Action action, FlowContext flowContext, Strin
...
+        if (versioningHandler == null) {
+            throw new ActionExecutionException("No action version handler found for action type: " + actionType);
+        }
+
+        try {
+            // If the action version is retired, skip executing the action.
+            if (versioningHandler.isRetiredActionVersion(actionType, action)) {
+                if (LOG.isDebugEnabled()) {
+                    LOG.debug(String.format("Action version is retired for action: %s.",  action.getId()));
+                }
```
ashanthamara: Let's add a diagnostic log as well for the retired action usages.

PR #7578, opened 2025-10-27
Adding case insensitive searching and sorting by API name to API Resource Management UI.

Jalina2007:
### Related Issue

[API resource filter is not working. #24875](https://github.com/wso2/product-is/issues/24875)

### Description
To handle the sorting issue, I changed the constants in the class SQLConstants.java which was enough to fix the sorting. This, however introduced another issue in the pagination logic, as that depended on the sorting to be done by CURSOR_KEY which the fix changed to be done by NAME. The pagination logic send the CURSOR_KEY of the last element of the current page to do the filtering at the SQL query, but since I changed the sorting to be done by NAME, sending the CURSOR_KEY does not make sense. So, I changed that logic in the PR [link](https://github.com/wso2/identity-api-server/pull/1029). This created another problem in the class FilterTreeBuilder.java, which checked pagination filters for the operator "pr". This worked earlier because CURSOR_KEY is an integer, but now since it contains text, parts of the NAME values got matched as operators. So I put a regex pattern there to make sure it matches that operator check only when it's at the end of the string. Other operators shouldn't cause this issue because they have spaces on both sides of the matching text. So, I left them untouched.

The fix for case insensitive sorting was straightforeward. I changed the SQL queries by adding the LOWER() function within the query in the file FilterQueriesUtil.java.

Before Change:
Screenshot 2025-10-27 114940
Screenshot 2025-10-27 115005

After Change:
Screenshot 2025-10-27 115330

Screenshot 2025-10-27 115350

### Follow up actions

I have run related tests in my local environment. If possible, please run any related tests to see if this breaks any other parts of the code, since I have applied changes to some core parts of the system. Please let me know if I have missed anything.

### Developer Checklist

- [x] [Behavioural Change] Does this change introduce a behavioral change to the product? 
- [ ] ↳ Approved by team lead
- [ ] ↳ Label impact/behavioral-change added
- [ ] ↳ Migration label added (e.g., 7.2.0-migration)
- [ ] ↳ Migration issues created and linked
- [ ] ↳ Label config added
- [ ] ↳ Configuration is properly documented

#### General

#### Functionality

#### Code

#### Tests

#### Security

#### Documentation

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search and filter functionality to support case-insensitive matching, enabling users to find results without needing to match exact capitalization.

### DIFF
--- a/components/api-resource-mgt/org.wso2.carbon.identity.api.resource.mgt/src/main/java/org/wso2/carbon/identity/api/resource/mgt/util/FilterQueriesUtil.java
+++ b/components/api-resource-mgt/org.wso2.carbon.identity.api.resource.mgt/src/main/java/org/wso2/carbon/identity/api/resource/mgt/util/FilterQueriesUtil.java
@@ -191,24 +191,27 @@ public class FilterQueriesUtil {
     private static void startWithFilterBuilder(int count, String value, String attributeName, StringBuilder filter,
                                                FilterQueryBuilder filterQueryBuilder) {
 
-        String filterString = " LIKE ? AND ";
-        filter.append(attributeName).append(filterString);
+        String filterString = " LIKE LOWER(?) AND ";
+        String attributeNameWithLowerCaseFunction = "LOWER(" + attributeName + ")";
+        filter.append(attributeNameWithLowerCaseFunction).append(filterString);
         filterQueryBuilder.setFilterAttributeValue(count, value + "%");
     }
 
     private static void endWithFilterBuilder(int count, String value, String attributeName, StringBuilder filter,
                                              FilterQueryBuilder filterQueryBuilder) {
 
-        String filterString = " LIKE ? AND ";
-        filter.append(attributeName).append(filterString);
+        String filterString = " LIKE LOWER(?) AND ";
+        String attributeNameWithLowerCaseFunction = "LOWER(" + attributeName + ")";
+        filter.append(attributeNameWithLowerCaseFunction).append(filterString);
         filterQueryBuilder.setFilterAttributeValue(count, "%" + value);
     }
 
     private static void containsFilterBuilder(int count, String value, String attributeName, StringBuilder filter,
                                               FilterQueryBuilder filterQueryBuilder) {
 
-        String filterString = " LIKE ? AND ";
-        filter.append(attributeName).append(filterString);
+        String filterString = " LIKE LOWER(?) AND ";
+        String attributeNameWithLowerCaseFunction = "LOWER(" + attributeName + ")";
+        filter.append(attributeNameWithLowerCaseFunction).append(filterString);
         filterQueryBuilder.setFilterAttributeValue(count, "%" + value + "%");
     }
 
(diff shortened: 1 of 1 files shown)


Malith-19:
@Jalina2007 Can you check the copilot suggestions and this filtering issue is only exists for the particular databases. In that case, normally we prepare the queries separately for those to avoid these kind of issues. Can we also look in to that possibility too?

Jalina2007:
Hi @Malith-19, I assume by the "filtering issue", you meant the main issue that we're addressing (case insensitive filtering). The problem is that some databases (MySQL, SQLite) are case-insensitive for the LIKE operator, but databases like PostgreSQL, Oracle, and H2 (which is the one used for testing) are case-sensitive for the same operator. The filter for the filtering is prepared for all platforms, and I found no reasonable way to "know" which database system is currently being used within the FilterQueriesUtil class (please correct me if I'm wrong). So the best way I found to fix this is to use explicit lowercasing of the values without depending on database-specific features.
I've also reviewed the Copilot suggestions and updated the code with the suggested improvements.
Please let me know if anything needs to be corrected

shashimalcse:
Hi @Jalina2007 did you test you solution on all the DB types?

Jalina2007:
Hi @shashimalcse , I ran the server on H2, MySQL, PostgreSQL, and OracleDB XE - everything worked as expected. However, when testing with Microsoft SQL Server XE, the server appeared to hang during startup. This might be a local environment issue on my machine. I’ll test it on another system and let you know.

Jalina2007:
Hi again, @shashimalcse , I managed to run the server with Microsoft SQL server and it worked as expected. The issue happened to be a configuration issue of the database side. Please let me know if I missed anything.

shashimalcse:
We need to find another way to handle ordering because your approach might cause pagination issues, and we need to think it through for other resources that use cursor-based pagination. Can we scope this PR to only address the search?

pavinduLakshan:
> We need to find another way to handle ordering because your approach might cause pagination issues, and we need to think it through for other resources that use cursor-based pagination. Can we scope this PR to only address the search?

@shashimalcse how about conditionally applying the resource name for the cursor when the configured DB is case-sensitive?

Jalina2007:
Hi @shashimalcse , I have removed the commits that tries to fix the ordering and narrowed the PRs scope to only fix the searching. Please let me know if anything needs to be changed.

shashimalcse:
Hi @Jalina2007 i still see the changes related to ordering in `components/api-resource-mgt/org.wso2.carbon.identity.api.resource.mgt/src/main/java/org/wso2/carbon/identity/api/resource/mgt/constant/SQLConstants.java` please check

Jalina2007:
Hi @shashimalcse , I'm sorry about the earlier confusion. I have updated it to only include the necessary commits for sorting.

Jalina2007:
Hi @shashimalcse  , Just checking in to see whether you need anything else from my side for the PR. Please let me know if I need to change anything. Thank you.

PR #7897, opened 2026-03-20
Support tenant-qualified URL routing for open APIs in sub-organizations

BimsaraBodaragama:
### Purpose

Support tenant-qualified sub-organization paths of the form `/t/{tenant-domain}/o/{org-id}/{api-path}` for open APIs, with correct sub-org tenant context resolution, proper validation, and clean error responses.

### PR Goals

- Add a dedicated `OrgContextsToRewriteInTenantQualifiedPaths` config block, decoupled from the existing `OrgContextsToRewriteInTenantPerspective`  (protected flow)
- Correctly set the sub-org's own tenant domain as the thread-local context  (not the root tenant domain) so requests behave identically to the existing  `/o/<org-id>/...` path
- Validate that the tenant domain in the URL matches the root tenant of the  given org, returning 404 for non-existent tenants and 400 for  tenant-org mismatches
- Handle unregistered `/t/.../o/.../` paths gracefully in `OrganizationResolver`  to prevent unhandled exceptions and spurious error logs
- Align the configured sub-paths with `OrgContextsToRewrite` to cover all  relevant branding-preference variants and validation-rules
- Ensure zero impact on existing org-context and tenant-perspective rewrite flows

### PR Approach

**`identity.xml.j2`**
- Introduced `<OrgContextsToRewriteInTenantQualifiedPaths>` block under `/api/`  base path with the following sub-paths:
  - `/api/server/v1/branding-preference`
  - `/api/server/v1/branding-preference/resolve`
  - `/api/server/v1/branding-preference/text`
  - `/api/server/v1/branding-preference/text/resolve`
  - `/api/server/v1/validation-rules`

**`OrganizationResolver.java`**
- Added `isTenantQualifiedOrgPath(String)` which reads `OrgContextsToRewriteInTenantQualifiedPaths.WebApp.Context.SubPaths.Path`  from `IdentityConfigParser` and short-circuits org resolution for registered paths, allowing `TenantContextRewriteValve` to handle them correctly
- Wrapped the `PATTERN_TENANT_AND_ORG_QUALIFIED` branch in a try-catch for `IdentityRuntimeException` and `RuntimeException` to gracefully handl  invalid tenant domains or org IDs on unregistered paths, preventing unhandled exceptions and server error logs

**`OrganizationResolverTest.java`**
- Added TestNG unit tests for `isTenantQualifiedOrgPath()` covering: exact matches, nested sub-paths, prefix-only non-matches (boundary fix), paths not starting with `/t/.../o/...`, blank URIs, empty config, and single-string config

### Related PRs
 - https://github.com/wso2-extensions/identity-carbon-auth-rest/pull/376
 - [next-sync] https://github.com/wso2/carbon-identity-framework/pull/8067

### Related Issues
 - https://github.com/wso2/product-is/issues/26822


components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/context/OrganizationResolver.java
```diff
@@ -229,6 +244,31 @@ private String getOrganizationTenantDomain(String organizationId) throws Organiz
...
+        }
+        List<String> subPaths = new ArrayList<>();
+        if (configuredPaths instanceof List) {
+            subPaths.addAll((List<String>) configuredPaths);
+        } else {
+            subPaths.add(configuredPaths.toString());
+        }
+        for (String subPath : subPaths) {
+            if (Pattern.compile("^/t/[^/]+/o/[^/]+" + Pattern.quote(subPath))
+                    .matcher(requestURI).find()) {
```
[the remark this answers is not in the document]
BimsaraBodaragama: Good catch. The regex was matching subpaths as prefixes, which could lead to unintended matches (e.g., `/validation-rules-v2`). Added a path boundary to ensure only the exact configured path (and valid nested paths) are matched.

Addressed in commit: [402ec98](https://github.com/wso2/carbon-identity-framework/pull/7897/commits/402ec98a7326e274c77a6bb3cedde3d82f187ba7)

components/identity-core/org.wso2.carbon.identity.core/src/test/java/org/wso2/carbon/identity/core/internal/context/OrganizationResolverTest.java
```diff
@@ -0,0 +1,152 @@
...
+                { "/t/acme.com/o/abc123/api/server/v1/offline-invite-link-copy", false },
+
+                // Paths that don't start with /t/.../o/... — should not match.
+                { "/api/server/v1/validation-rules", false },
+                { "/o/abc123/api/server/v1/validation-rules", false },
+
+                // Blank / null URI.
+                { "", false },
+                { "   ", false },
+
```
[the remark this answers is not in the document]
BimsaraBodaragama: Null is handled upstream in `resolveOrganizationInContext` via StringUtils.isBlank, so it never reaches this logic. Updated the comment to reflect the actual coverage.

Addressed in commit: [f58e13a](https://github.com/wso2/carbon-identity-framework/pull/7897/commits/f58e13af49fdc539883b73a7844ce00976437c3e)

BimsaraBodaragama:
### ✅ Fix Verification

https://github.com/user-attachments/assets/b4ba0e45-a066-4038-813f-14c911bd5566

components/identity-core/org.wso2.carbon.identity.core/src/main/java/org/wso2/carbon/identity/core/internal/context/OrganizationResolver.java
```diff
@@ -103,6 +97,29 @@ public void resolveOrganizationInContext(Request request) {
...
+        } catch (IdentityRuntimeException e) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Invalid tenant domain in request URI: " + requestURI, e);
+            }
+        } catch (RuntimeException e) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Error resolving organization context for request URI: " + requestURI, e);
+            }
+        } catch (OrganizationManagementException | UserStoreException e) {
+            LOG.error("Error while initializing organization information.", e);
```
SujanSanjula96, line 120: Shall we check what happen when /o/<org-id> open API paths with incorrect org ids

BimsaraBodaragama: Previously, when `/o/<invalid-org-id>/api/server/v1/...` was called with a non-existent org-id, `resolveRootAndSubOrganization()` would call `getPrimaryOrganizationId()` first which silently returned the super org ID as a fallback for unknown org-ids. This caused `carbon.super` to be set as the root organization on `IdentityContext`, and `resolveOrganization()` would then silently return null (org not found). The result was that the API handler operated under the `carbon.super` context and returned the root tenant's data instead of an error. For example, `validation-rules` would return `carbon.super`'s validation rules with a 200 OK.
>
> The fix is in two places:
>
> 1. **`OrganizationContextRewriteValve`**: added `isValidOrganization()` which calls `organizationManager.isOrganizationExistById()` before setting the org-id on the Carbon context. If the org doesn't exist, the valve now returns a 400 response immediately, preventing the request from reaching the API handler with the wrong context.
> Addressed in commit: [b90f0d9](https://github.com/wso2-extensions/identity-carbon-auth-rest/pull/376/commits/b90f0d911e1b14a90bb524be7e4a86256aa8565e)
>
> 2. **`OrganizationResolver.resolveRootAndSubOrganization()`**: added an upfront `getMinimalOrganization()` check before resolving the root org. If the org doesn't exist or the lookup throws a `RuntimeException` (due to null tenant context at that pipeline stage), the method returns early without setting anything on `IdentityContext`, preventing the silent fallback to `carbon.super` context.
> Addressed in commit: [d69db4b](https://github.com/wso2/carbon-identity-framework/pull/7897/commits/d69db4bd3fc24acc8266ae94f0955c31a1606291)
>
> Both fixes together ensure that an invalid org-id on `/o/` paths consistently returns a 400 with no server-side errors.

BimsaraBodaragama:
## ✅ Fix Verification

https://github.com/user-attachments/assets/62de2184-de36-47ff-9a6d-474bc94b28a4

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
```diff
@@ -4321,6 +4321,21 @@
         </Servlet>
     </OrgContextsToRewriteInTenantPerspective>
 
+    <OrgContextsToRewriteInTenantQualifiedPaths>
```
SujanSanjula96:
```suggestion
    <OrgResourceContextsToRewriteInTenantPerspective>
```

BimsaraBodaragama: Addressed in commit: [d4d2320](https://github.com/wso2/carbon-identity-framework/pull/7897/commits/d4d2320029f29d2e5efd31083736565b4f401a7f)

BimsaraBodaragama:
## ✅ Fix verification

https://github.com/user-attachments/assets/aef67231-bcea-4a3f-998c-caf0849098cb

SujanSanjula96 (approved):

BimsaraBodaragama:
## ✅ Final Fix Verification

https://github.com/user-attachments/assets/98763377-27a6-4dff-b12b-3e0a8bcbc405

PR #7983, opened 2026-04-06
Fix: Deprecated handle method in JIT Provisioning flow breaks customizations extended from DefaultProvisioningHandler

Zeta201:
Related issue: https://github.com/wso2/product-is/issues/27454

### Problem
When v2 roles were introduced, the `handleWithV2Roles` method in the JIT provisioning flow did not expose the `AuthenticationContext` to the provisioning handler. This broke customizations where deployers extended `DefaultProvisioningHandler` to intercept and transform raw IDP roles before applying them to the local user.

Without access to AuthenticationContext, custom provisioning handlers had no way to retrieve the original claims sent by the federated IDP, making it impossible to replicate the role-mapping logic.

### Root Cause
`DefaultStepBasedSequenceHandler` called `handleWithV2Roles(roleIdList, subject, attributes, userStoreDomain, tenantDomain)` without passing the `AuthenticationContext`. As a result, custom subclasses of `DefaultProvisioningHandler` had no access to the authentication context (and therefore the raw IDP claims) when the method was invoked.

### Fix
- Added a new overloaded `handleWithV2Roles` method to the `ProvisioningHandler` interface that accepts `AuthenticationContext` as an additional parameter. 
- A default implementation is provided that delegates to the existing 5-argument method, ensuring full backward compatibility for existing custom implementations.
- Overrode this new method in DefaultProvisioningHandler, delegating to the existing implementation.
- Updated `DefaultStepBasedSequenceHandler` to call the new 6-argument overload, passing the `AuthenticationContext` through to the handler.

### Impact
Custom provisioning handlers extending `DefaultProvisioningHandler` can now override `handleWithV2Roles(List, String, Map, String, String, AuthenticationContext)` to access the full authentication context, including IDP-provided claims, to implement custom role-mapping logic. Existing implementations that do not override the new signature are unaffected due to the default delegation.

### DIFF
--- a/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/provisioning/ProvisioningHandler.java
+++ b/components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/handler/provisioning/ProvisioningHandler.java
@@ -18,6 +18,7 @@
 
 package org.wso2.carbon.identity.application.authentication.framework.handler.provisioning;
 
+import org.wso2.carbon.identity.application.authentication.framework.context.AuthenticationContext;
 import org.wso2.carbon.identity.application.authentication.framework.exception.FrameworkException;
 
 import java.util.List;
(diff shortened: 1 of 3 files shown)


Zeta201:
Closing this PR since the changes are already being tracked in https://github.com/wso2/carbon-identity-framework/pull/8070

PR #8138, opened 2026-06-04
Add cache for application authorized scopes

### LINKED ISSUE #27984: Redundant Database Calls on Cached Token Retrieval When Scopes Are Requested in Client Credentials Grant
### Description

Invoking client credentials grant with scopes executes an additional API call to get authorized scopes when an active access token is retrieved from the cache

If the client credentials grant is executed with a scope requested, the `AuthorizedAPIDAO::getAuthorizedScopes` [1] method will be executed each time a token call is received by WSO2 IS. This will execute an additional database call since no cache is available for authoriuzed scopes.

When using opaque access tokens, subsequent token calls after the first token call (which generates a new access token, saves it in the database, and adds it to the cache) will be retrieved from the cache without invoking any database calls.

However, if the token call has a scope requested, `getAuthorizedScopes` will execute a query every time. This did not happen in previous versions such as IS 5.11.0.

Deployments with high token request counts will see an impact on the TPS of token calls and could experience database issues as a result.

The authorized scopes for the application must be cached instead of querying the database each time or skip retrieving the authorized scopes from the database when the token is active and available in the cache.

[1] - https://github.com/wso2/carbon-identity-framework/blob/c7c8fa8beee8a46e3fe8a23e98c01db241b08471/components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/AuthorizedAPIManagementServiceImpl.java#L228
[2] - https://github.com/wso2/carbon-identity-framework/blob/c7c8fa8beee8a46e3fe8a23e98c01db241b08471/components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/dao/impl/CacheBackedAuthorizedAPIDAOImpl.java#L83-L87

### Steps to Reproduce

1. Create an application with opaque access tokens
2. Authorize and API resource
3. Iniate a client credential grant token request with the scope
4. Check the database calls
5. Initiate the same token request before the token cache is expired. This will return the same access token.
6. Check the database calls
7. Execute the client credentials grant without any scopes
8. Execute the same request again before the cache is expired. The same access token will be returned.
9. No database calls will be executed

### Please select the area issue is related to

API Access Mgt & Authorization

### Version

IS 7.2.0

### Environment Details (with versions)

_No response_

### Developer Checklist

- [ ] [Behavioural Change] Does this change introduce a behavioral change to the product?
- [ ]  ↳ Approved by team lead
- [ ]  ↳ Label `impact/behavioral-change` added
- [ ]  ↳ Migration label added (e.g., `7.2.0-migration`)
- [ ]  ↳ Migration issues created and linked
- [ ]  ↳ Label `config` added
- [ ]  ↳ Configuration is properly documented

shashimalcse:
## Purpose

Fixes wso2/product-is#27984

When a client-credentials token is requested **with a scope**, `AuthorizedAPIDAO.getAuthorizedScopes(appId, tenantId)` runs a DB query on **every** token request — even when the access token itself is served from the token cache — because the app-level authorized-scopes lookup has no cache layer. Under high token TPS this becomes a database bottleneck (a regression vs. older releases).

## Approach

Add a per-application authorized-scopes cache, mirroring the existing `getAuthorizedAPI` read-through pattern in `CacheBackedAuthorizedAPIDAOImpl` and reusing the standard `BaseCache` infrastructure (tenant partitioning + cluster invalidation).

- New `AuthorizedScopesCache` / `AuthorizedScopesCacheKey` (keyed by `appId`) / `AuthorizedScopesCacheEntry`.
- `getAuthorizedScopes` now reads through the cache (empty results cached too, to avoid re-querying apps with no authorized scopes).
- **Invalidation** on every app-level mutation: both `addAuthorizedAPI` overloads, `patchAuthorizedAPI`, and `deleteAuthorizedAPI`.
- **Cross-component invalidation:** a new `AuthorizedScopesCacheInvalidationHandler` (`AbstractEventHandler`) flushes the cache on API Resource Management events (`POST_DELETE_API_RESOURCE`, `POST_DELETE_API_RESOURCE_SCOPES`, `POST_DELETE_SCOPE`, `POST_PUT_API_RESOURCE_SCOPES`, `POST_UPDATE_API_RESOURCE`), since those mutations do not flow through the DAO.

## Notes

- The invalidation handler must be subscribed via the `identity-event` configuration to fire; a default subscription should be added in the product distribution so API-resource/scope-deletion invalidation works out of the box. The DAO-mutator invalidations are always active.
- Behaviour is unchanged functionally (same scopes returned) — the cache only removes the redundant per-request DB lookup.

## Testing

- Verified at runtime (equivalent change on the 7.2.0 update line): repeat token requests served from cache (no extra `GET_AUTHORIZED_SCOPES` query); cache correctly invalidated on authorized-API mutations and on API-resource scope deletion.


ThaminduR (dismissed):

components/api-resource-mgt/org.wso2.carbon.identity.api.resource.mgt/src/main/java/org/wso2/carbon/identity/api/resource/mgt/dao/impl/CacheBackedAPIResourceMgtDAO.java
```diff
@@ -273,34 +285,65 @@ public Scope getScopeByNameTenantIdAPIId(String name, Integer tenantId, String a
...
             throws APIResourceMgtException {
 
-        return apiResourceManagementDAO.getScopesByTenantId(tenantId, expressionNodes);
+        // Only the unfiltered "all tenant scopes" lookup is cached (the hot path used on the OAuth token
+        // flow when authorize_all_scopes is enabled). Filtered lookups pass through to avoid a key explosion.
+        if (expressionNodes != null && !expressionNodes.isEmpty()) {
+            return apiResourceManagementDAO.getScopesByTenantId(tenantId, expressionNodes);
+        }
+
+        ScopeMetadataCacheKey cacheKey =
```
piraveena, line 294: Why we are giving this constant value as the cache key? Shouldn't the key be tenantDomain or tenantID?

PasinduYeshan:
https://github.com/wso2/carbon-identity-framework/pull/8166

PR #8147, opened 2026-06-10
Fix returning immutable objects from the service level in APIAuthorization and RoleManagement

ashanthamara:
- Returning immutable objects can be problematic, as there can be requirements to apply modifications to those objects in the Listener level. Hence remove the immutability options from the ApplicationAPIAuthorization and Role Manangement

### Related issue
- https://github.com/wso2/product-is/issues/28023


shashimalcse (approved):

ashanthamara:
Integration test runner passed in the fork:
- https://github.com/ashanthamara/product-is/actions/runs/27269950493

PR #8149, opened 2026-06-11
Add proxy host support to the External API Framework

DilshanSenarath:
## Purpose

Adds optional HTTP proxy support to the External API Framework client. Callers can now configure a proxy host and port via `APIClientConfig.Builder`, and the underlying Apache HttpClient will route all requests through the specified proxy.

### `APIClientConfig`
- Added `proxyHost` (String) and `proxyPort` (int) fields with defaults of `null` and `0`.
- Exposed `getProxyHost()` and `getProxyPort()` getters.
- Added `proxyHost(String)` and `proxyPort(int)` builder methods.

### `APIClient`
- After building the `HttpClientBuilder`, checks `StringUtils.isNotBlank(proxyHost)` and calls `clientBuilder.setProxy(new HttpHost(host, port))` only when a proxy is configured — no behaviour change for existing callers.

### Tests
- **`APIClientConfigTest`** — 5 new unit tests covering: builder stores proxy host/port, default values are `null`/`0`, and explicit `null` is preserved.
- **`APIClientTest`** — 2 new integration tests:
  - Verifies that an unreachable proxy causes the call to fail even when the target server is up (proves the proxy path is taken).
  - Verifies that a blank proxy host is ignored and the direct connection succeeds.

## Related Issue
- https://github.com/wso2/product-is/issues/28017

### DIFF
--- a/components/external-api-framework/org.wso2.carbon.identity.external.api.client/src/main/java/org/wso2/carbon/identity/external/api/client/api/model/APIClientConfig.java
+++ b/components/external-api-framework/org.wso2.carbon.identity.external.api.client/src/main/java/org/wso2/carbon/identity/external/api/client/api/model/APIClientConfig.java
@@ -34,6 +34,8 @@ public class APIClientConfig {
     private final int poolSizeToBeSet;
     private final int maxPerRoute;
     private final long responseLimitInBytes;
+    private final String proxyHost;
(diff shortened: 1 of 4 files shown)


PasinduYeshan (approved):

DilshanSenarath:

PR #8153, opened 2026-06-12
Add support for closing the HTTP client

DilshanSenarath:
## Purpose

`APIClient` creates a `PoolingHttpClientConnectionManager` and a `CloseableHttpClient` during construction. Both are long-lived, stateful resources that hold pooled TCP connections. Neither was ever released, meaning implementations that extend `AbstractAPIClientManager` had no way to clean up the underlying HTTP client on shutdown, leading to a resource leak.

### `APIClient`
- Added a `close()` method that calls `httpClient.close()`, which in turn shuts down the underlying `PoolingHttpClientConnectionManager` and releases all pooled connections.

### `AbstractAPIClientManager`
- Added a `close()` method that delegates to `apiClient.close()`.
- Downstream components that extend this class should call `close()`.

### Tests
- Updated `APIClientTest.tearDown()` to call `apiClient.close()` after each test for proper cleanup.
- Added `testCloseSucceeds` — verifies `close()` on a fresh `APIClient` does not throw.
- Added `testCloseAfterSuccessfulRequestSucceeds` — makes a real HTTP call then closes, verifying no exception is thrown after the client has been used.
- Added `AbstractAPIClientManagerTest` covering:
  - `testCloseSucceeds` — verifies `close()` on a fresh manager does not throw.
  - `testCloseAfterSuccessfulRequestSucceeds` — verifies close after serving a request.
  - `testCallAPIDelegatesSuccessfully` — verifies the abstract class correctly delegates to the underlying `APIClient`.
  - `testCallAPIWithNullArgumentsThrowsIllegalArgumentException` — covers the null guard in `callAPI()`.

## Related Issue
- https://github.com/wso2/product-is/issues/28017

### DIFF
--- a/components/external-api-framework/org.wso2.carbon.identity.external.api.client/src/main/java/org/wso2/carbon/identity/external/api/client/api/service/AbstractAPIClientManager.java
+++ b/components/external-api-framework/org.wso2.carbon.identity.external.api.client/src/main/java/org/wso2/carbon/identity/external/api/client/api/service/AbstractAPIClientManager.java
@@ -27,6 +27,8 @@ import org.wso2.carbon.identity.external.api.client.api.model.APIRequestContext;
 import org.wso2.carbon.identity.external.api.client.api.model.APIResponse;
 import org.wso2.carbon.identity.external.api.client.internal.service.APIClient;
 
+import java.io.IOException;
+
 /**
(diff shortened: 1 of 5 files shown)


ZiyamSanthosh (approved):

DilshanSenarath:

PR #8169, opened 2026-06-19
Merge same policy scopes into a single authorized scope entry

DilshanSenarath:
### Purpose
When `authorize_all_scopes = true` and `authorize_internal_scopes = false`, the `AuthorizedAPIManagementServiceImpl.getAuthorizedScopes()` method returns two separate `AuthorizedScopes` entries both carrying `policyId = RBAC`:

1. An all-tenant entry (all non-internal scopes, including server level default scopes (Ex:- `default`))
2. The app's API resource subscription entry (only the subscribed scopes)

The `DefaultOAuth2ScopeValidator` iterates over these entries and stores handler results in a map keyed by handler name. Since both entries are handled by `M2MScopeValidationHandler` (same policyId), the second result overwrites the first, and scopes like `"default"` that are absent from the app's subscriptions are dropped.

The fix is to merge the app's internal/console scopes into the single all-tenant RBAC entry instead of appending a separate RBAC entry — ensuring the handler processes only one entry and the overwrite never occurs.

**Config combination behaviour:**

| `authorize_all_scopes` | `authorize_internal_scopes` | App has subscriptions | Result |
|---|---|---|---|
| `false` | any | any | Only app's subscribed scopes returned — no issue |
| `true` | `true` | any | Single all-tenant entry (internal scopes included) — no issue |
| `true` | `false` | no | Single filtered all-tenant entry — no issue |
| `true` | `false` | yes | **Two RBAC entries → second overwrites first → scopes like `"default"` dropped** |

To control this behavior, a new configuration `merge_authorized_scopes_by_policy` will be introduced:

```toml
[oauth]
merge_authorized_scopes_by_policy = true
```

When enabled, authorized scope entries sharing the same `policyId` are merged into a single entry before validation, preventing the overwrite. **This will be enabled by default**, as it represents the correct behavior aligned with the intent of `authorize_all_scopes = true`. If an user reports a regression expecting the previous behavior (apps with subscriptions restricted to only their subscribed scopes), they can be guided to explicitly set `merge_authorized_scopes_by_policy = false` to restore the old behavior.

### Related Issue
- https://github.com/wso2/product-is/issues/27644

### DIFF
--- a/components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/ApplicationConstants.java
+++ b/components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/ApplicationConstants.java
@@ -149,6 +149,7 @@ public class ApplicationConstants {
     public static final String MYACCOUNT_PORTAL_PATH = "MyAccount.AppBaseName";
     public static final String AUTHORIZE_ALL_SCOPES = "OAuth.AuthorizeAllScopes";
     public static final String AUTHORIZE_INTERNAL_SCOPES = "OAuth.AuthorizeInternalScopes";
+    public static final String MERGE_AUTHORIZED_SCOPES_BY_POLICY = "OAuth.MergeAuthorizedScopesByPolicy";
     public static final String RBAC = "RBAC";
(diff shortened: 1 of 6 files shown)


pavinduLakshan (dismissed):

pavinduLakshan (approved):

DilshanSenarath:

PR #8171, opened 2026-06-21
Fix NullPointerException in LoggerUtils.getSanitizedErrorMessage when userName is null

AmshikaH:
## Summary

- `LoggerUtils.getSanitizedErrorMessage(String errorMessage, String userName)` calls `errorMessage.contains(userName)` without a null-check on `userName`
- When the refresh token grant fails after a user password reset, the user context is unresolved and `userName` is `null` — `String.contains(null)` throws NPE, surfacing as HTTP 500 `server_error`
- Fix: add `userName != null &&` guard before `errorMessage.contains(userName)`

## Related Issues

- https://github.com/wso2/product-is/issues/27853


AfraHussaindeen (dismissed):

AmshikaH:
The above builds failed due to the timeout issue. The successful build triggered without the timeout can be found below.

PR builder completed
https://github.com/AmshikaH/wso2-product-is/actions/runs/28214308515
Status: success

AfraHussaindeen (approved):

PR #8178, opened 2026-06-23
Add granular scope feature

mpmadhavig:
## Purpose

Resolve console feature action scopes (e.g. `console:applications_edit`, `console:applications_create`) as **holders** when they appear inside another collection's `<Read>` / `<Create>` / `<Update>` / `<Delete>` block. The holder is replaced by the leaf scopes from the matching action block of the owning collection, and the holder name itself is dropped from the bucket — so the `apiResources` payload of `/api-resource-collections/{id}` surfaces the underlying internal scopes (and their owning API resource) instead of an opaque console scope that no API resource owns.

## Changes

- **Parser** (`APIResourceCollectionMgtConfigBuilder`): pre-build a `holder → leaf scopes` map from each collection's `<Feature>` block + matching action block. `_view` → owner Read, `_create`/`_update`/`_delete` → matching owner block, `_edit` → owner Create + Update + Delete (legacy coarse write). Resolution is recursive with cycle protection + memoisation so an owner's action block can itself reference another holder.
- **`api-resource-collection.xml.j2`**: declare `_create`/`_update`/`_delete` siblings next to every `_edit` holder inside non-`<Feature>` blocks (8 sites: `applicationAuthenticationScript`, `applicationClientSecretManagement`, `applicationInternalAPIAuthorization`, `userSharingV2`, and the four `org_*` counterparts).
- **Tests**: 7 new cases in `APIResourceCollectionMgtConfigBuilderTest` — update/view/edit holder resolution, write-bucket propagation, transitive recursion, literal scopes preserved alongside holders, owner collection unaffected.

## Related Issues
- https://github.com/wso2/product-is/issues/27974

### DIFF
--- a/components/api-resource-mgt/org.wso2.carbon.identity.api.resource.collection.mgt/src/main/java/org/wso2/carbon/identity/api/resource/collection/mgt/util/APIResourceCollectionMgtConfigBuilder.java
+++ b/components/api-resource-mgt/org.wso2.carbon.identity.api.resource.collection.mgt/src/main/java/org/wso2/carbon/identity/api/resource/collection/mgt/util/APIResourceCollectionMgtConfigBuilder.java
@@ -127,12 +127,16 @@ public class APIResourceCollectionMgtConfigBuilder {
             if (scopesElement != null) {
                 Set<String> readScopeSet = new HashSet<>();
                 Set<String> writeScopeSet = new HashSet<>();
+                Set<String> createScopeSet = new HashSet<>();
+                Set<String> updateScopeSet = new HashSet<>();
+                Set<String> deleteScopeSet = new HashSet<>();
                 Iterator<?> actionElements = scopesElement.getChildElements();
                 while (actionElements.hasNext()) {
                     OMElement actionElement = (OMElement) actionElements.next();
                     if (actionElement == null) {
                         continue;
                     }
+                    String actionName = actionElement.getLocalName();
                     Iterator<OMElement> scopes = actionElement.getChildrenWithName(
                             new QName(APIResourceCollectionConfigBuilderConstants.SCOPE_ELEMENT));
                     while (scopes.hasNext()) {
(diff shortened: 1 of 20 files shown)


components/api-resource-mgt/org.wso2.carbon.identity.api.resource.collection.mgt/src/main/java/org/wso2/carbon/identity/api/resource/collection/mgt/util/APIResourceCollectionMgtConfigBuilder.java
```diff
@@ -127,41 +130,99 @@ private void buildAPIResourceCollectionConfig() {
...
+                                    updateScopeSet.add(scopeName);
+                                } else if (isDeleteAction) {
+                                    deleteScopeSet.add(scopeName);
+                                }
                             }
                         } else {
                             // Process new scopes with feature scopes.
                             if (isReadAction) {
-                                readScopeSet.add(scopeName);
+                                if (isHolderScope(scopeName)) {
```
mpmadhavig: isHolderScope logic need to be added only for `non-COLLECTION_VERSION_V0`.

mpmadhavig: This logic will flatten the scopes out.

hwupathum: Shall we use a different method name? Its hard to understand what it does without looking at the method itself

mpmadhavig: Good point. Renamed `isHolderScope` to `isFeatureScope`, which matches the `isViewFeatureScope`/`isEditFeatureScope`/etc. helpers it delegates to and reads clearly at the call sites.

mpmadhavig: Resolved dc414aa7ec46fcbe7ad45a024316aed8275620e9

hwupathum (approved):

PR #8201, opened 2026-07-10
Expose the fields that caused a rule to fail in the rule evaluation result

kaviska:
## Proposed changes in this pull request

This PR adds the ability to know **which fields caused a rule to fail** during evaluation, and exposes them on the `RuleEvaluationResult`.

Previously, `RuleEvaluationResult` only reported **whether** a rule was satisfied (a boolean). Callers had no way to tell **why** a rule did not match. This PR adds a `failedFields` list so callers can see the specific field names that failed, which is useful for diagnostics, user feedback, and debugging.

The change is confined to the rule evaluation module and is additive and backward compatible:

- **`RuleEvaluationResult`** — add a `failedFields` list, a constructor that accepts it, and a `getFailedFields()` accessor. The existing constructor is kept (it delegates with an empty list), and the accessor returns an unmodifiable, never-null list (empty when the rule is satisfied).
- **`RuleEvaluator`** — while evaluating the OR-of-ANDs rule tree, collect the field names of the failing expressions per OR branch. If a branch passes, the collected failed fields are cleared (the rule is satisfied, so nothing failed); if all branches fail, the failing fields from the branches are retained. Exposed via `getFailedFields()`.
- **`RuleEvaluationServiceImpl`** — pass the evaluator's failed fields into the returned `RuleEvaluationResult`.

The component provides:

- The list of fields responsible for a rule not matching, in addition to the existing pass/fail result.
- Backward compatibility: existing callers that use the boolean result and the old constructor are unaffected. `failedFields` is an additive side output and does not change the evaluation outcome.

Unit tests assert that the failed fields are reported when a rule fails and are empty when the rule passes.

## When should this PR be merged

No preconditions.

## Follow up actions

N/A

### DIFF
--- a/components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
+++ b/components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
@@ -18,6 +18,9 @@
 
 package org.wso2.carbon.identity.rule.evaluation.api.model;
 
+import java.util.Collections;
+import java.util.List;
+
 /**
  * Rule evaluation result.
  * This class is used to represent the result of a rule evaluation.
@@ -26,11 +29,18 @@ public class RuleEvaluationResult {
 
     private final String ruleId;
     private final boolean ruleSatisfied;
+    private final List<String> failedFields;
 
     public RuleEvaluationResult(String ruleId, boolean ruleSatisfied) {
 
+        this(ruleId, ruleSatisfied, Collections.emptyList());
+    }
+
+    public RuleEvaluationResult(String ruleId, boolean ruleSatisfied, List<String> failedFields) {
+
         this.ruleId = ruleId;
         this.ruleSatisfied = ruleSatisfied;
+        this.failedFields = failedFields != null ? failedFields : Collections.emptyList();
     }
 
     public String getRuleId() {
@@ -41,5 +51,14 @@ public class RuleEvaluationResult {
     public boolean isRuleSatisfied() {
 
         return ruleSatisfied;
+    }
+
+    /**
+     * Returns the list of fields that failed evaluation.
+     * Empty when the rule is satisfied.
+     */
+    public List<String> getFailedFields() {
+
+        return Collections.unmodifiableList(failedFields);
     }
 }
--- a/components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
+++ b/components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
@@ -48,7 +48,6 @@ public class RuleEvaluationResult {
 
         this.ruleId = ruleId;
         this.ruleSatisfied = ruleSatisfied;
-        // Keep as-is: null means the failed fields were not set, which is different from an empty list.
         this.failedFields = failedFields;
     }
 
--- a/components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
+++ b/components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
@@ -31,16 +31,25 @@ public class RuleEvaluationResult {
     private final boolean ruleSatisfied;
     private final List<String> failedFields;
 
+    /**
+     * Creates a result without setting the failed fields.
+     *
+     * @deprecated Use {@link #RuleEvaluationResult(String, boolean, List)} which also captures the failed
+     *             fields. This constructor leaves the failed fields unset ({@code null}), so callers cannot
+     *             distinguish "no fields failed" from "failed fields were not computed".
+     */
+    @Deprecated
     public RuleEvaluationResult(String ruleId, boolean ruleSatisfied) {
 
-        this(ruleId, ruleSatisfied, Collections.emptyList());
+        this(ruleId, ruleSatisfied, null);
     }
 
     public RuleEvaluationResult(String ruleId, boolean ruleSatisfied, List<String> failedFields) {
 
         this.ruleId = ruleId;
         this.ruleSatisfied = ruleSatisfied;
-        this.failedFields = failedFields != null ? failedFields : Collections.emptyList();
+        // Keep as-is: null means the failed fields were not set, which is different from an empty list.
+        this.failedFields = failedFields;
     }
 
     public String getRuleId() {
@@ -54,11 +63,13 @@ public class RuleEvaluationResult {
     }
 
     /**
-     * Returns the list of fields that failed evaluation.
+     * Returns the list of fields that failed evaluation, or {@code null} if the failed fields were not set.
      * Empty when the rule is satisfied.
+     *
+     * @return Unmodifiable list of failed field names, or {@code null} if not set.
      */
     public List<String> getFailedFields() {
 
-        return Collections.unmodifiableList(failedFields);
+        return failedFields == null ? null : Collections.unmodifiableList(failedFields);
     }
 }
--- a/components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/internal/service/impl/RuleEvaluator.java
+++ b/components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/internal/service/impl/RuleEvaluator.java
@@ -70,25 +70,24 @@ public class RuleEvaluator {
      */
     public boolean evaluate(Rule rule, Map<String, FieldValue> evaluationData) throws RuleEvaluationException {
 
-        return evaluate(null, rule, evaluationData).isRuleSatisfied();
+        return evaluateResult(rule, evaluationData).isRuleSatisfied();
     }
 
     /**
      * Evaluate a given rule and return the result, including the fields that failed evaluation.
      *
-     * @param ruleId         Rule ID.
      * @param rule           Rule to evaluate.
      * @param evaluationData Evaluation data.
      * @return Rule evaluation result with the satisfied status and the failed fields.
      * @throws RuleEvaluationException If an error occurs while evaluating the rule.
      */
-    public RuleEvaluationResult evaluate(String ruleId, Rule rule, Map<String, FieldValue> evaluationData)
+    public RuleEvaluationResult evaluateResult(Rule rule, Map<String, FieldValue> evaluationData)
             throws RuleEvaluationException {
 
         List<String> failedFields = new ArrayList<>();
         ORCombinedRule orRule = (ORCombinedRule) rule;
         boolean ruleSatisfied = evaluateORCombinedRule(orRule, evaluationData, failedFields);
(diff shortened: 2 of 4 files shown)


components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
```diff
@@ -26,11 +29,18 @@ public class RuleEvaluationResult {
 
     private final String ruleId;
     private final boolean ruleSatisfied;
+    private final List<String> failedFields;
 
     public RuleEvaluationResult(String ruleId, boolean ruleSatisfied) {
 
+        this(ruleId, ruleSatisfied, Collections.emptyList());
```
Thisara-Welmilla: Keep this null, So we can identify that this is value is not being set.

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
```diff
@@ -26,11 +29,18 @@ public class RuleEvaluationResult {
 
     private final String ruleId;
     private final boolean ruleSatisfied;
+    private final List<String> failedFields;
 
     public RuleEvaluationResult(String ruleId, boolean ruleSatisfied) {
```
Thisara-Welmilla, line 42: Make this method deprecated

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/internal/service/impl/RuleEvaluator.java
```diff
@@ -46,6 +48,7 @@ public class RuleEvaluator {
     private static final Log LOG = LogFactory.getLog(RuleEvaluator.class);
 
     private final OperatorRegistry operatorRegistry;
+    private final List<String> failedFields = new ArrayList<>();
```
Thisara-Welmilla: This is incorrect. This class is a singleton and concurrently executing the evaluate, therefore we cannot have such attribute as this.

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/internal/service/impl/RuleEvaluator.java
```diff
@@ -71,22 +74,38 @@ public boolean evaluate(Rule rule, Map<String, FieldValue> evaluationData) throw
         return evaluateORCombinedRule(orRule, evaluationData);
     }
 
+    /**
+     * Returns the list of fields that failed evaluation for the last evaluated rule.
+     * Empty when the rule is satisfied.
+     *
+     * @return List of failed field names.
+     */
+    public List<String> getFailedFields() {
```
Thisara-Welmilla: Add a new evaluate overloaded method to return the RuleEvaluationResult. So the calling class can access the result from returning object.

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/internal/service/impl/RuleEvaluator.java
```diff
@@ -71,22 +74,38 @@ public boolean evaluate(Rule rule, Map<String, FieldValue> evaluationData) throw
...
+
+        return Collections.unmodifiableList(failedFields);
+    }
+
     private boolean evaluateORCombinedRule(ORCombinedRule orRule, Map<String, FieldValue> evaluationData)
             throws RuleEvaluationException {
 
         for (ANDCombinedRule andRule : orRule.getRules()) {
-            if (evaluateANDCombinedRule(andRule, evaluationData)) {
+            List<String> branchFailedFields = new ArrayList<>();
```
Thisara-Welmilla, line 84: Shouldn't this go out of the for loop, as we need to keep all failed fields?

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
```diff
@@ -26,11 +29,18 @@ public class RuleEvaluationResult {
 
     private final String ruleId;
     private final boolean ruleSatisfied;
+    private final List<String> failedFields;
```
Thisara-Welmilla, line 32: I am thinking, whether this should be nested, to build meaningful results message to the user. Can you brainstorm on this?

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
kaviska: addressed in commit https://github.com/wso2/carbon-identity-framework/pull/8201/changes/7e1b00f7b6a0c8a10e74bea190f828bd566eca1d

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
kaviska: https://github.com/wso2/carbon-identity-framework/pull/8201/changes/7e1b00f7b6a0c8a10e74bea190f828bd566eca1d

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/internal/service/impl/RuleEvaluator.java
kaviska: To add an overloaded evaluate method, the parameter list (or type) has to differ . So I added ruleId as a new parameter, which is also needed to construct the RuleEvaluationResult. This is what the current implementation does.
The alternative would be to change the existing public evaluate(Rule, Map) method to return the result directly (deriving the id from rule.getId()), but changing/removing a public method breaks existing callers and isn't a good practice — so I preferred keeping the existing method intact and adding the overload.

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/internal/service/impl/RuleEvaluator.java
kaviska: because of this line it not apply here right
 **failedFields.addAll(branchFailedFields);**

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/internal/service/impl/RuleEvaluator.java
Thisara-Welmilla: Lets have new method name with same params

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
kaviska: we can try like this format

`[
  [   // first branch (index 0)
     { field: "osVersion",   operator: ">=", expected: "12", actual: "10" },
     { field: "deviceType",  operator: "==", expected: "phone", actual: "tablet" }
  ],
  [   // second branch (index 1)
     { field: "rooted",      operator: "==", expected: "false", actual: "true" }
  ]
]`

But my opinion is it's look like overhead for the consumer side and these information do not need for the consumer

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/internal/service/impl/RuleEvaluator.java
Thisara-Welmilla: This a internal class, and it is meant to be used by only within the component. If so, it is okay to update the public method. @kaviska check whether this is exposed expose to out of component from the pom.xml

kaviska: ` <Export-Package>
                            !org.wso2.carbon.identity.rule.evaluation.internal.*,
                            org.wso2.carbon.identity.rule.evaluation.api.*;
                            version="${carbon.identity.package.export.version}"
                        </Export-Package>`
                        
All internal class are not exported shall I remove the old method ?

kaviska: I have change evaluate method

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
```diff
@@ -26,11 +29,27 @@ public class RuleEvaluationResult {
...
 
+    /**
+     * Creates a result without setting the failed fields.
+     *
+     * @deprecated Use {@link #RuleEvaluationResult(String, boolean, List)} which also captures the failed
+     *             fields. This constructor leaves the failed fields unset ({@code null}), so callers cannot
+     *             distinguish "no fields failed" from "failed fields were not computed".
+     */
+    @Deprecated
```
kaviska, line 41: The repo has other  @Deprecated annotations and none use since/forRemoval, so adding them only here would be inconsistent

components/rule-mgt/org.wso2.carbon.identity.rule.evaluation/src/main/java/org/wso2/carbon/identity/rule/evaluation/api/model/RuleEvaluationResult.java
```diff
@@ -26,11 +29,27 @@ public class RuleEvaluationResult {
...
     public RuleEvaluationResult(String ruleId, boolean ruleSatisfied) {
 
+        this(ruleId, ruleSatisfied, null);
+    }
+
+    public RuleEvaluationResult(String ruleId, boolean ruleSatisfied, List<String> failedFields) {
+
         this.ruleId = ruleId;
         this.ruleSatisfied = ruleSatisfied;
+        // Keep as-is: null means the failed fields were not set, which is different from an empty list.
```
Thisara-Welmilla: This comment should go to the line 44. right?

kaviska: The comment was removed because java doc explain it

Thisara-Welmilla (approved):

PR #8228, opened 2026-07-28
Introduce the device policy component

kaviska:
## Proposed changes in this pull request

This PR introduces a new component, **`org.wso2.carbon.identity.device.policy`**, which decides whether a device is allowed to authenticate. It takes a signed **device token**, proves the token came from a registered device and has not been replayed, enriches the device attributes with platform-verified integrity data, and evaluates those attributes against a device compliance policy.

This is the **consuming component** in the device management set. It depends on `org.wso2.carbon.identity.device.mgt` for the device registry and on the policy management and policy evaluation components for the policy model and rule evaluation. It publishes services consumed by the adaptive authentication engine, the device registration flow, and the device policy REST API.

The change adds the component under the `components/device-mgt` aggregator, registers that aggregator in the root POM, and adds the device token replay table to the database scripts.

### Database

One table is added to all supported database scripts (DB2, H2, MS SQL, MySQL, MySQL Cluster, Oracle, Oracle RAC, PostgreSQL):

- **`IDN_DEVICE_TOKEN_JTI`** — the device token replay store: the token's `jti`, tenant, issued-at time and expiry, keyed on `(TENANT_ID, JTI)` so a given token identifier can be spent exactly once per tenant.
- One supporting index, `IDX_IDN_DEVICE_TOKEN_JTI_EXPIRY`, used by the cleanup task that removes expired rows.

### Device token verification

Device data is not accepted as a plain payload. It arrives as a JWT signed by the device's private key and is verified before any attribute is trusted. `DeviceTokenExtractor` performs the full check in one place:

- **Signature** — the `deviceId` custom header parameter resolves the device in the registry, and the token's ECDSA signature is verified against that device's registered public key. A device that is absent or not `ACTIVE` is rejected before any signature work.
- **Freshness** — the `iat` claim must be within a five minute window, with a thirty second allowance for a device clock running ahead of the server. A token older or further ahead than that is rejected.
- **Single use** — the `jti` claim is checked against the replay store and recorded there, so a captured token cannot be presented twice. The stored row expires at `iat` plus the freshness window, which is exactly as long as the token could still be replayed.

There are two entry points, because the two callers establish trust in the public key differently. `extractFromToken(...)` looks the device up in the registry and uses its stored key, and is what authentication uses. `extractWithPublicKey(...)` takes the key from the caller and performs no registry lookup — the registration flow uses it, because at that point the device is not yet registered but the flow has just proven possession of the key via the challenge signature. Both paths enforce signature, freshness and single-use identically.

### The component provides

- **`DevicePolicyEvaluator`** — the main OSGi service contract. `evaluate(policyName, deviceData, appId, tenantDomain)` returns `null` when the device is compliant, and otherwise a comma-separated list of the fields that caused the failure, so the caller can tell the user what to fix rather than only that the check failed. Missing required fields are reported before evaluation is attempted, and a policy that cannot be resolved is reported distinctly from one that was evaluated and failed.
- **`DeviceTokenVerifier`** — verification for callers that already hold the device's public key. Consumed by the device registration component.
- **`DeviceFieldMetadataService`** — exposes which device fields apply to which platforms, driven by `device-fields.json`. Consumed by the device policy REST API layer so the console can show only the fields relevant to a chosen platform.
- **A `DeviceDataResolver` implementation** — resolves the device token at authentication initiation, reading it from the `device_token` query parameter for redirect flows and falling back to the `X-Device-Token` header for app-native and headless flows. A token that fails verification yields no device data rather than an error, leaving the policy check to decide the outcome.
- **`isDevicePolicyCompliant`** — an adaptive authentication JS function registered with the `JsFunctionRegistry`, so a login script can branch on device compliance. It reads the already-verified device data from the authentication context rather than re-verifying the token per script call.
- **A `RuleEvaluationDataProvider`** for the `DEVICE_POLICY` flow, converting device attributes into typed rule field values and rejecting a non-numeric value for a numeric field rather than silently coercing it.
- **Integrity enrichment** — `attestationToken` is exchanged with the client attestation service for a platform-verified integrity verdict, and is removed from the device data before evaluation so a policy rule never sees the raw token. Android verdicts are reduced to the strongest level the device meets; iOS resolves to an attested flag. When attestation cannot be performed the fields are set to their failed values rather than left absent, so a policy cannot pass by omission.
- **A scheduled cleanup task** that removes expired `jti` rows, modelled on the framework's pushed authorization request cleanup. It is enabled by default on a fifteen minute period, configurable through `JDBCPersistenceManager.DeviceTokenJtiCleanUp` in `identity.xml`, and unlike the existing framework cleanup services it shuts its scheduler down on bundle deactivation and survives a failed run rather than cancelling the schedule.
- **A `DevicePolicyErrorMessage` catalog** with stable `DPM-` error codes, surfaced through the policy management client and server exception types, plus diagnostic logging at the token validation, integrity enrichment and policy evaluation choke points.

### Consumers

- The **adaptive authentication engine** calls the `isDevicePolicyCompliant` JS function from login scripts, and uses the `DeviceDataResolver` service to capture verified device data at initiation.
- The **device registration component** calls `DeviceTokenVerifier` to verify the device payload during registration, and `DevicePolicyEvaluator` to enforce a registration-time policy before the device record is written.
- The **device policy REST API** (in the `identity-api-server` repository) calls `DeviceFieldMetadataService` to build the platform-aware field metadata response.

### Testing

The component ships a TestNG suite with JaCoCo enforcing 80% line and 60% complexity coverage at the bundle level, with no exclusions. Coverage focuses on the token verification path — signature failure, missing and malformed `deviceId`, inactive and unknown devices, undecodable keys, missing `jti` and `iat`, stale and future-dated tokens including the clock-skew and freshness boundaries, and replay rejection — asserting the specific `DPM-` error code in each case rather than only the exception type. The replay DAO is tested against H2 including tenant isolation, and the policy evaluator, integrity enricher, JS function, data resolver and configuration loader are covered against their branches.

## When should this PR be merged

This PR has preconditions and should be merged **last** of the device management set. It does not build until the following are available:

- **`org.wso2.carbon.identity.device.mgt`** — the device registry, used to resolve a device and its public key.
- **`org.wso2.carbon.identity.policy.management`** and **`org.wso2.carbon.identity.policy.evaluation`** — the policy model and evaluation engine.
- **Rule management `getFailedFields` support** — the evaluator reports which rule fields failed, which requires the failed-fields addition to `RuleResourceEvaluationResult`.

Building the branch today fails on exactly those three unresolved artifacts and nothing else, which is the expected state until they reach the Maven repository.

## Follow up actions

- **P2 feature packaging** — a `features/device-mgt` server feature is needed to ship the bundle into the IS distribution and to deploy `device-fields.json`, which the component requires at activation. Not included here, matching the device management component PR.
- **Rule metadata** — the `DEVICE_POLICY` flow definition and its device field definitions in the rule metadata resources, which belong with the rule management change that introduces the flow type.
- **Device policy REST API** in the `identity-api-server` repository.
- **Audit logging** for policy evaluation outcomes, currently covered by diagnostic logging only.


components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DeviceFieldMetadataService.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+public interface DeviceFieldMetadataService {
+
+    /**
+     * Returns a map of field name to the list of platforms the field applies to.
+     * Fields that apply to all platforms are omitted from the map.
+     *
+     * @return map of field name -> applicable platforms.
+     */
+    Map<String, List<String>> getFieldApplicablePlatforms();
```
Thisara-Welmilla: We need to have exception type for this component and throw that exception here.

Thisara-Welmilla: Where are we using this method? I am thinking whether it should be map of platform -> field names

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
```diff
@@ -0,0 +1,55 @@
...
+     * @param policyName   Name of the policy to evaluate against.
+     * @param deviceData   Mutable map of device field names to their values.
+     * @param appId        App resource ID for loading credentials.
+     * @param tenantDomain Tenant domain for policy lookup and evaluation.
+     * @return {@code null} if compliant, or a failure reason string if not.
+     * @throws PolicyManagementException If the policy cannot be retrieved.
+     * @throws RuleEvaluationException   If rule evaluation fails.
+     * @throws PolicyEvaluationException If policy evaluation fails.
+     */
+    String evaluate(String policyName, Map<String, Object> deviceData,
```
Thisara-Welmilla: Instead of policyName, lets use policyId. From the JS function it need to resolve the policyId from policyname and invoke this method.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/cleanup/DeviceTokenJtiCleanupService.java
```diff
@@ -0,0 +1,92 @@
...
+    private final long initialDelay;
+    private final long delayBetweenRuns;
+
+    /**
+     * Constructor for DeviceTokenJtiCleanupService.
+     *
+     * @param initialDelay     Initial delay in minutes before the first run.
+     * @param delayBetweenRuns Delay between runs in minutes.
+     */
+    public DeviceTokenJtiCleanupService(long initialDelay, long delayBetweenRuns) {
```
Thisara-Welmilla, line 47: Do we have such other cleanup running in the IS runtime. AFAIK, normally we are doing db clean procedures.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/config/DeviceFieldConfig.java
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * Represents a single device field entry from device-fields.json.
+ * Captures the field name, the HTTP header it maps to, and the platforms it applies to.
+ * Fields with no applicablePlatforms are universal (shown for all platforms).
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class DeviceFieldConfig {
+
+    private String name;
+    private List<String> applicablePlatforms;
```
Thisara-Welmilla: How about other way around. Have platform as enum and under them field names as property

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DeviceTokenReplayService.java
```diff
@@ -0,0 +1,94 @@
...
+import org.wso2.carbon.identity.policy.management.api.exception.PolicyManagementServerException;
+
+import java.sql.Timestamp;
+import java.util.Date;
+
+/**
+ * Enforces single-use of device token jti claims against the replay store, and removes expired
+ * jti records. This is the only class in the bundle that constructs {@link DeviceTokenReplayDAOImpl}.
+ */
+public class DeviceTokenReplayService {
```
Thisara-Welmilla: Name must have `protection` or `prevent` something

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/IntegrityDataEnricherImpl.java
```diff
@@ -0,0 +1,110 @@
...
+    private static final String IOS_INTEGRITY_KEY = "iosIntegrity";
+    private static final String INTEGRITY_FAILED = "INTEGRITY_FAILED";
+
+    private final DevicePolicyDiagnosticLogger diagnosticLogger = new DevicePolicyDiagnosticLogger();
+
+    @Override
+    public void enrich(Map<String, Object> deviceData, String appId, String tenantDomain) {
+
+        String attestationToken = (String) deviceData.remove(ATTESTATION_TOKEN_KEY);
+        if (attestationToken == null) {
```
Thisara-Welmilla: For now, lets make attestationToken validation mandatory. Later we can bring UI toggle to config based on that we can allow here. If attestationToken is null, we allow, this can be misused.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
Thisara-Welmilla: What are the deviceData have? I feel this should be a model defining from this class. So the whoever call this method, can build that object.

Thisara-Welmilla: Response is also need to properly modeled.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/IntegrityDataEnricherImpl.java
```diff
@@ -0,0 +1,110 @@
...
+                    "Attestation token verification failed.");
+            LOG.error("Failed to verify attestation token for device policy enrichment "
+                    + "in application: " + appId + ", tenant: " + tenantDomain, e);
+            deviceData.put(ANDROID_INTEGRITY_KEY, INTEGRITY_FAILED);
+            deviceData.put(IOS_INTEGRITY_KEY, String.valueOf(false));
+        }
+    }
+
+    // Priority: MEETS_STRONG_INTEGRITY > MEETS_DEVICE_INTEGRITY > MEETS_BASIC_INTEGRITY > MEETS_VIRTUAL_INTEGRITY.
+    private String resolveAndroidIntegrityLevel(List<String> verdicts) {
```
Thisara-Welmilla: Have two separate path for platform specific config and separate code logic from file level for android, ios, etc.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/test/resources/repository/resources/identity/device-policy/device-fields.json
```diff
@@ -0,0 +1,12 @@
+{
+  "fields": [
+    {
+      "name": "device_model",
```
Thisara-Welmilla: Lets make these enum/constant in code level, and avoid introduce new config json files.

features/device-mgt/org.wso2.carbon.identity.device.mgt.server.feature/resources/identity/device-policy/device-fields.json
```diff
@@ -0,0 +1,90 @@
+{
+  "fields": [
```
Thisara-Welmilla: Lets bring these to code level, in to device Field class. If we see a requirement to config these, we can bring capability through the identity.xml

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/js/DevicePolicyJsFunction.java
```diff
@@ -0,0 +1,83 @@
...
+import java.util.Map;
+import java.util.function.BiFunction;
+
+/**
+ * JS function for device policy compliance check, registered as {@code isDevicePolicyCompliant}.
+ * The verified device payload is captured and stored on the authentication context at initiation
+ * (see {@code DeviceDataResolverImpl}); this function reads it from the context under
+ * {@link FrameworkConstants#DEVICE_DATA}.
+ */
+public class DevicePolicyJsFunction implements BiFunction<JsBaseAuthenticationContext, String, String> {
```
Thisara-Welmilla, line 43: Normal we keep these functions in https://github.com/wso2-extensions/identity-conditional-auth-functions/tree/master/components. Let move this there

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/util/DeviceTokenExtractor.java
```diff
@@ -0,0 +1,236 @@
...
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Extracts and verifies device attribute data from a signed device JWT.
+ * The JWT header must contain a deviceId custom parameter used to look up the registered public key.
+ * Signature is verified using ECDSA before claims are returned.
+ */
+public class DeviceTokenExtractor {
```
Thisara-Welmilla, line 56: Bring this to the util package

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/cleanup/DeviceTokenJtiCleanupService.java
kaviska: We have these cleanup services 

PushedAuthReqCleanupService, SessionCleanUpService,

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
kaviska: This function used in two place js function and the executor . without name to id to conversation in two places my opinion is we can keep this in this method

kaviska: Android and iOS send different field sets, so a model with a member per field would be two-thirds null on any request

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/js/DevicePolicyJsFunction.java
kaviska: Shall we keep this here , there are some conditional auth function like push progressive  enrolment  are not in there also if we moving this function we need import policy device policy component as well

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/IntegrityDataEnricherImpl.java
kaviska: if we do mandatory app need to register in google and apple do the process even for the os version check even in this method we ignore this DevicePolicyEvaluator we have check to see all the files in policy should come in the request .So in policy if admin configure attersataion and if in app not send token it will failed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/util/DeviceTokenExtractor.java
kaviska: addressed

features/device-mgt/org.wso2.carbon.identity.device.mgt.server.feature/resources/identity/device-policy/device-fields.json
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/test/resources/repository/resources/identity/device-policy/device-fields.json
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/IntegrityDataEnricherImpl.java
kaviska: for this resolve happened only in android

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DeviceTokenReplayService.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/config/DeviceFieldConfig.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DeviceFieldMetadataService.java
kaviska: addressed

kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/constant/DevicePolicyErrorMessage.java
```diff
@@ -0,0 +1,116 @@
...
+
+package org.wso2.carbon.identity.device.policy.api.constant;
+
+/**
+ * Error messages for the Device Policy component.
+ */
+public enum DevicePolicyErrorMessage {
+
+    // Client errors.
+    ERROR_DEVICE_TOKEN_PARSE_FAILED("DPM-60008", "Device token invalid.",
```
Thisara-Welmilla: Why we are staring with 60008. Also 60010 is missing

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DeviceFieldMetadataService.java
```diff
@@ -0,0 +1,51 @@
...
+import org.wso2.carbon.identity.device.policy.api.exception.DevicePolicyServerException;
+import org.wso2.carbon.identity.device.policy.api.model.Platform;
+
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Provides platform applicability metadata for device policy fields.
+ */
+public interface DeviceFieldMetadataService {
```
Thisara-Welmilla: DeviceFieldMetadataService -> DevicePolicyMetadataService. Keep this generic

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DeviceTokenVerifier.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Map;
+
+/**
+ * Verifies a device-data JWT against a caller-supplied public key and returns
+ * the verified claims. Used where trust in the key has already been
+ * established out of band (e.g. the device registration executor, which
+ * verifies the key via the challenge signature) and the device is not yet
+ * persisted, so a registry lookup is neither possible nor required.
+ */
+public interface DeviceTokenVerifier {
```
Thisara-Welmilla: Use more generic class name, instead of specific `verifier` name.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DeviceTokenVerifier.java
```diff
@@ -0,0 +1,49 @@
...
+     * single-use checks before the claims are trusted.
+     *
+     * @param token           Raw JWT string carrying the device attributes.
+     * @param base64PublicKey Base64-encoded X.509 EC public key to verify.
+     * @param correlationId   Identifier used only for diagnostic correlation.
+     * @param tenantDomain    Tenant domain used for replay-store scoping.
+     * @return Map of device field names to their values from verified JWT.
+     * @throws DevicePolicyException If the token is invalid or stale.
+     */
+    Map<String, Object> verifyWithPublicKey(String token, String base64PublicKey,
```
Thisara-Welmilla: Lets have a more meaning full name, like resolveAndVerifyDataFromToken

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DeviceFieldMetadataServiceImpl.java
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Implementation of DeviceFieldMetadataService backed by the DeviceField enum.
+ */
+public class DeviceFieldMetadataServiceImpl implements DeviceFieldMetadataService {
+
+    @Override
+    public List<String> getFieldsForPlatform(Platform platform) throws DevicePolicyServerException {
+
+        List<String> fields = new ArrayList<>();
```
Thisara-Welmilla: As this list is static, this should not be building at runtime, we can have contant, here or may be in model class. Think whether it fits best.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DeviceFieldMetadataServiceImpl.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.List;
+import java.util.Set;
+
+/**
+ * Implementation of DeviceFieldMetadataService backed by the DeviceField enum.
+ */
+public class DeviceFieldMetadataServiceImpl implements DeviceFieldMetadataService {
```
Thisara-Welmilla: I think we can get rid of this service and use the model enum class to provide the same capabilities.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
```diff
@@ -0,0 +1,51 @@
...
+     * themselves.
+     *
+     * @param policyName   Name of the policy to evaluate against.
+     * @param deviceData   Mutable map of device field names to their values.
+     * @param appId        App resource ID for loading credentials.
+     * @param tenantDomain Tenant domain for policy lookup and evaluation.
+     * @return A {@link DevicePolicyEvaluationResult} representing the evaluation status and field details.
+     * @throws DevicePolicyException If an error occurs during policy evaluation.
+     */
+    DevicePolicyEvaluationResult evaluate(String policyName, Map<String, Object> deviceData,
```
Thisara-Welmilla, line 48: Shall we have policyId instaed of policyName?

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
```diff
@@ -0,0 +1,161 @@
...
+
+        String platform = (String) deviceData.get(DEVICE_PLATFORM_FIELD);
+
+        List<String> missingFields = findMissingRequiredFields(policyName, platform, deviceData, tenantDomain);
+        if (!missingFields.isEmpty()) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("Device data incomplete for policy '" + policyName + "': " + missingFields);
+            }
+            return DevicePolicyEvaluationResult.incompleteDeviceData(policyName, missingFields);
+        }
```
Thisara-Welmilla: Move this part also to the `findMissingRequiredFields` method and rename to a proper method name

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
```diff
@@ -0,0 +1,161 @@
...
+        try {
+            result = DevicePolicyComponentServiceHolder.getInstance()
+                    .getPolicyEvaluationService()
+                    .evaluate(policyId, platform != null ? platform : "", context, tenantDomain);
+        } catch (PolicyEvaluationException e) {
+            throw DevicePolicyExceptionHandler.handleServerException(
+                    DevicePolicyErrorMessage.ERROR_DEVICE_POLICY_EVALUATION_FAILED, e, policyName);
+        }
+
+        if (result == null) {
```
Thisara-Welmilla: How results can be null?

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
```diff
@@ -0,0 +1,161 @@
...
+                    .collect(Collectors.toList());
+            return DevicePolicyEvaluationResult.nonCompliant(policyName, failedFields);
+        }
+        return DevicePolicyEvaluationResult.compliant(policyName);
+    }
+
+    private List<String> findMissingRequiredFields(String policyName, String platform,
+            Map<String, Object> deviceData, String tenantDomain) throws DevicePolicyServerException {
+
+        if (platform == null || platform.trim().isEmpty()) {
```
Thisara-Welmilla: We can use StringUtil.isBlank instead of `platform == null || platform.trim().isEmpty())`. isn't it?

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
```diff
@@ -0,0 +1,161 @@
...
+            return Collections.singletonList(DEVICE_PLATFORM_FIELD);
+        }
+        Policy policy;
+        try {
+            policy = DevicePolicyComponentServiceHolder.getInstance()
```
Thisara-Welmilla: This check should go to the DevicePolicyComponentServiceHolder.getInstance().getPolicyEvaluationService().evaluate. So each place who call this method does not need check these.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
```diff
@@ -0,0 +1,161 @@
...
+                    .getPolicyManagementService()
+                    .getPolicyByName(policyName, tenantDomain);
+        } catch (PolicyManagementException e) {
+            throw DevicePolicyExceptionHandler.handleServerException(
+                    DevicePolicyErrorMessage.ERROR_DEVICE_POLICY_EVALUATION_FAILED, e, policyName);
+        }
+        if (policy == null) {
+            return Collections.emptyList();
```
Thisara-Welmilla: Why not throwing an error?

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/IntegrityDataEnricherImpl.java
Thisara-Welmilla: We need to have config from UI, to enable, disable per policy or per platform.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/IntegrityDataEnricherImpl.java
```diff
@@ -0,0 +1,110 @@
...
+                deviceData.put(IOS_INTEGRITY_KEY, String.valueOf(ctx.isAttested()));
+            }
+
+        } catch (ClientAttestationMgtException e) {
+            diagnosticLogger.logIntegrityEnrichmentFailure(appId,
+                    "Attestation token verification failed.");
+            LOG.error("Failed to verify attestation token for device policy enrichment "
+                    + "in application: " + appId + ", tenant: " + tenantDomain, e);
+            deviceData.put(ANDROID_INTEGRITY_KEY, INTEGRITY_FAILED);
+            deviceData.put(IOS_INTEGRITY_KEY, String.valueOf(false));
```
Thisara-Welmilla: At this point, if an exception occurred,  we need to break the flow.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/IntegrityDataEnricher.java
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Enriches the device data map with platform-verified integrity values before policy evaluation.
+ *
+ * <p>Handles both Android and iOS platforms via the underlying client attestation service, which
+ * detects the platform automatically from the token format (JWE for Android, CBOR for iOS).
+ * Removes {@code attestationToken} from the map before returning — policy rules never see the raw token.
+ * Sets {@code androidIntegrity} for Android devices and {@code iosIntegrity} for iOS devices.
+ * When the client attestation service is not available, both fields are set to their failed values.
+ */
+public interface IntegrityDataEnricher {
```
Thisara-Welmilla: I do not this we need a interface for this.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/util/DeviceTokenExtractor.java
```diff
@@ -0,0 +1,234 @@
...
+     * @throws DevicePolicyException If the token is invalid or signature verification fails.
+     */
+    public Map<String, Object> extractFromToken(String token, String tenantDomain)
+            throws DevicePolicyException {
+
+        try {
+            SignedJWT signedJWT = parseSignedJWT(token);
+
+            String deviceId = (String) signedJWT.getHeader().getCustomParam(DEVICE_ID_PARAM);
+            if (deviceId == null || deviceId.trim().isEmpty()) {
```
Thisara-Welmilla: Use StringUtil

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/util/DeviceTokenExtractor.java
```diff
@@ -0,0 +1,234 @@
...
+                diagnosticLogger.logTokenValidationFailure(null, "Device token header is missing the deviceId.");
+                throw DevicePolicyExceptionHandler.handleClientException(
+                        DevicePolicyErrorMessage.ERROR_DEVICE_TOKEN_MISSING_DEVICE_ID);
+            }
+            // Strip any stray leading/trailing quotes or backslashes introduced by JWT serialisation bugs.
+            deviceId = deviceId.replaceAll("[\"\\\\]+$", "").replaceAll("^[\"\\\\]+", "").trim();
```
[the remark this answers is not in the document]
Thisara-Welmilla: @kaviska please check on this.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/util/DeviceTokenExtractor.java
```diff
@@ -0,0 +1,234 @@
...
+
+            Device device = DevicePolicyComponentServiceHolder.getInstance()
+                    .getDeviceManagementService()
+                    .getDeviceById(deviceId, tenantDomain);
+
+            if (device == null || device.getStatus() != Device.Status.ACTIVE) {
```
Thisara-Welmilla, line 93: Purpose this method and class is extract token, This check should go the evaluate method.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DeviceFieldMetadataService.java
kaviska: DevicePolicyMetadataService would collide with identity-api-server's existing DevicePolicyMetadataService  two classes with the same name across the repo boundary, which caused a runtime NoSuchMethodError  Going with DevicePolicyFieldMetadataService instead

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
kaviska: This function used in inside the js function and the executor without covert this in two places it's better to do conversation inside this method ?

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/util/DeviceTokenExtractor.java
kaviska: But to exact public key that public key should active. that's why add that validation here

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/constant/DevicePolicyErrorMessage.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DeviceTokenVerifier.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DeviceTokenVerifier.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DeviceFieldMetadataServiceImpl.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DeviceFieldMetadataServiceImpl.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
kaviska: addressed.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/IntegrityDataEnricherImpl.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/IntegrityDataEnricher.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/util/DeviceTokenExtractor.java
kaviska: addressed

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/model/DevicePolicyEvaluationResult.java
```diff
@@ -0,0 +1,151 @@
...
+public class DevicePolicyEvaluationResult {
+
+    /**
+     * Evaluation outcome status.
+     */
+    public enum Status {
+        COMPLIANT,
+        NON_COMPLIANT,
+        INCOMPLETE_DEVICE_DATA,
+        POLICY_NOT_FOUND
```
Thisara-Welmilla: Why do you think this should be a status `POLICY_NOT_FOUND`. Isn't it error scenario, where service should throw an error.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/model/DeviceField.java
```diff
@@ -0,0 +1,169 @@
...
+
+/**
+ * Defines device policy fields and the platforms they apply to.
+ */
+public enum DeviceField {
+
+    PLATFORM("platform", EnumSet.allOf(Platform.class)),
+    LOCK_SCREEN("lockScreen", EnumSet.allOf(Platform.class)),
+
+    ANDROID_OS_VERSION("androidOsVersion", EnumSet.of(Platform.ANDROID)),
```
Thisara-Welmilla, line 38: I think, it is better to field as reference in the platform, than refering platform here. It is the natural way of being used; eg: get field for a platform.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
```diff
@@ -0,0 +1,51 @@
...
+                                           String appId, String tenantDomain)
```
Thisara-Welmilla, line 49: What is the app Id refering here?

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DeviceTokenService.java
```diff
@@ -0,0 +1,49 @@
...
+     * single-use checks before the claims are trusted.
+     *
+     * @param token           Raw JWT string carrying the device attributes.
+     * @param base64PublicKey Base64-encoded X.509 EC public key to verify.
+     * @param correlationId   Identifier used only for diagnostic correlation.
+     * @param tenantDomain    Tenant domain used for replay-store scoping.
+     * @return Map of device field names to their values from verified JWT.
+     * @throws DevicePolicyException If the token is invalid or stale.
+     */
+    Map<String, Object> resolveAndVerifyDataFromToken(String token, String base64PublicKey,
```
Thisara-Welmilla, line 46: This is a part of evaluation process. Why  do you think of having separate class for this, instead of having it in DevicePolicyEvaluator?

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
Thisara-Welmilla: In the flow execution, it is the policy Id. Isn't it? As the policy id is the unique identifier, better to use that

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/resolver/DeviceDataResolverImpl.java
```diff
@@ -0,0 +1,91 @@
...
+    public Optional<Map<String, Object>> resolveDeviceData(HttpServletRequest request, String tenantDomain) {
+
+        if (request == null) {
+            return Optional.empty();
+        }
+
+        String deviceToken = request.getParameter(DEVICE_TOKEN_PARAM);
+        if (StringUtils.isBlank(deviceToken)) {
+            deviceToken = request.getHeader(DEVICE_TOKEN_HEADER);
+        }
```
Thisara-Welmilla, line 70: Incorrect logic. Remove this and add `deviceToken = request.getHeader(DEVICE_TOKEN_HEADER);` to  line 74

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/resolver/DeviceDataResolverImpl.java
```diff
@@ -0,0 +1,91 @@
...
+        if (StringUtils.isBlank(deviceToken)) {
+            return Optional.empty();
+        }
+
+        try {
+            Map<String, Object> deviceData = new DeviceTokenExtractor().extractFromToken(deviceToken, tenantDomain);
+
+            String attestationHeader = request.getHeader(Constants.ATTESTATION_HEADER);
+            if (StringUtils.isNotBlank(attestationHeader)) {
```
Thisara-Welmilla, line 79: Attestation verification happen here.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
```diff
@@ -0,0 +1,160 @@
...
+    private static final String FLOW_TYPE_DEVICE_POLICY = "DEVICE_POLICY";
+
+    @Override
+    public DevicePolicyEvaluationResult evaluate(String policyName, Map<String, Object> deviceData,
+                                                 String appId, String tenantDomain)
+            throws DevicePolicyException {
+
+        DevicePolicyComponentServiceHolder.getInstance()
+                .getIntegrityDataEnricher()
+                .enrich(deviceData, appId, tenantDomain);
```
Thisara-Welmilla, line 63: Why this comes here. It belong data resolving part. isn't it?

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
```diff
@@ -0,0 +1,160 @@
...
+            throw DevicePolicyExceptionHandler.handleServerException(
+                    DevicePolicyErrorMessage.ERROR_DEVICE_POLICY_EVALUATION_FAILED, e, policyName);
+        }
+    }
+
+    private Optional<DevicePolicyEvaluationResult> checkDeviceDataCompleteness(String policyName, String platform,
+            Policy policy, Map<String, Object> deviceData) {
+
+        List<String> missingFields;
+        if (StringUtils.isBlank(platform)) {
```
Thisara-Welmilla, line 120: Before the check, it has been used in line 66 `String platform = (String) deviceData.get(DEVICE_PLATFORM_FIELD);`

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
```diff
@@ -0,0 +1,160 @@
...
+            missingFields = Collections.singletonList(DEVICE_PLATFORM_FIELD);
+        } else {
+            missingFields = findMissingRequiredFields(platform, policy, deviceData);
```
Thisara-Welmilla, line 123: To policy be success, only part of expression is satisfying enough. So, even with. partial data policy can be success and customer might write the policy in that way also.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
kaviska: for attestation is configured for the per app

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/resolver/DeviceDataResolverImpl.java
kaviska: In the login path if client send in that header attention already available in the police evelaaution we read the result. but in the register path client send attestation in jwt so in attestation context no attention result

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
kaviska: https://github.com/wso2/carbon-identity-framework/pull/8228#discussion_r3687444982

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
kaviska: yes, but for the policy evaluation the sdk need to send the fields configured for the policy

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DeviceTokenService.java
kaviska: In the registration path he not need to look db to get public key , because he already have one . This used in registration  path

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/resolver/DeviceDataResolverImpl.java
kaviska: since as written the current logic looks correct to me: try the device_token query param first, fall back to the X-Device-Token header if blank, and only bail out if neither is present. That matches the fallback behavior documented in the class Javadoc (query param for redirect flows, header for native/headless flows).

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
kaviska: In the Flow execution we keep policy name . we keep this name in executor metadata property not in specific column  also I thing in the login path we use name for evaluation it's better to keep consistency

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/constant/DevicePolicySQLConstants.java
```diff
@@ -0,0 +1,65 @@
...
+    public static final class Query {
+
+        public static final String IS_TOKEN_REPLAYED =
+                "SELECT 1 FROM IDN_DEVICE_TOKEN_JTI " +
+                        "WHERE JTI = :JTI; AND TENANT_ID = :TENANT_ID;";
+
+        public static final String STORE_TOKEN_JTI =
+                "INSERT INTO IDN_DEVICE_TOKEN_JTI " +
+                        "(JTI, TENANT_ID, IAT, EXPIRY_TIME) " +
+                        "VALUES (:JTI;, :TENANT_ID;, :IAT;, :EXPIRY_TIME;)";
```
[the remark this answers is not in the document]
kaviska: This isn't malformed SQL — the ; here isn't a statement terminator, it's the placeholder delimiter required by org.wso2.carbon.database.utils.jdbc.NamedPreparedStatement, which this codebase uses for named-parameter SQL instead of raw ? binding. I decompiled the library and confirmed its tokenizer scans for : to start a parameter and ; to end it, throwing "Cannot find the end of the placeholder." if the ; is missing — so removing it would break these queries at runtime, not fix them.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/model/DeviceField.java
Malith-19: If we move the declaration to the platform wise level, then for the common fields there will be duplications right? In this way, each field is declared once and linked the supporting OS directly, which increase the maintainability in the future.

Malith-19: Regarding the getting the fields per OS is already present and `getFieldNamesForPlatform` method will return the field list when we pass the OS to the method.

Malith-19, line 38: Shall we improve the grouping of the fields? eg: group all the versioning fields together.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
Malith-19: @Thisara-Welmilla Are we good to go with the policyName here?

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/cleanup/DeviceTokenJtiCleanupService.java
Malith-19: Is it possible to disable this service and use a DB cleanup procedure as well?

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/cleanup/DeviceTokenJtiCleanupService.java
```diff
@@ -0,0 +1,92 @@
...
+
+        scheduler.shutdownNow();
+    }
+
+    /**
+     * Device token jti cleanup task.
+     */
+    private static final class DeviceTokenJtiCleanupTask implements Runnable {
+
+        private final DeviceTokenReplayProtectionService replayService =
```
Malith-19, line 76:
```suggestion
        private final DeviceTokenReplayProtectionService replayProtectionService =
```

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json
```diff
@@ -52,6 +52,8 @@
   "session.timeout.maximum_session_timeout": "30d",
   "par.cleanup.enable_expired_requests_cleanup": true,
   "par.cleanup.clean_expired_requests_every": "1d",
+  "device_token.cleanup.enable_expired_jti_cleanup": true,
+  "device_token.cleanup.clean_expired_jti_every": 15,
```
Malith-19, line 56: is this 15mins cleanups? Also for the par cleanup time we are using string ("1d") and in the device token we are using directly an integer which limits us to only have the time in minutes. 

@Thisara-Welmilla shall we update this to use the string notation and do we need the 15 cleanup as default?

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/model/DevicePolicyEvaluationResult.java
Malith-19: Removed Already.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/model/DeviceField.java
Malith-19: Fixed.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/cleanup/DeviceTokenJtiCleanupService.java
Malith-19: This cleanup can be disabled from the deployment.toml via configuration. Hence closing this.

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json
Thisara-Welmilla: Lets go with 1D

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/cleanup/DeviceTokenJtiCleanupService.java
Malith-19: Fixed with https://github.com/wso2/carbon-identity-framework/pull/8228/commits/09e9020a48696d632f2fe347613772b7c3ab716d

features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/org.wso2.carbon.identity.core.server.feature.default.json
Malith-19: Fixed with https://github.com/wso2/carbon-identity-framework/pull/8228/commits/dcc0f71fd80270edd4255ccd5d36e256cbe61e5b

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
Malith-19: We can go with the policyId and give the policyName support as well but it should be using this evaluate method once it derived the policyId from the given policyName

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/IntegrityDataEnricherImpl.java
Malith-19: We need to introduce the UI config and when we introducing set the attestation enable by default.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/js/DevicePolicyJsFunction.java
Malith-19: Need to check if we have already moved this or unless we have to move it as thisara suggested

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/model/DeviceField.java
```diff
@@ -0,0 +1,175 @@
...
+     *
+     * @param platform Platform to look up, or {@code null} for every known field.
+     * @return List of applicable field wire names.
+     */
+    public static List<String> getFieldNamesForPlatform(Platform platform) {
+
+        if (platform == null) {
+            return ALL_FIELD_NAMES;
+        }
+        return FIELDS_BY_PLATFORM.getOrDefault(platform, Collections.emptyList());
```
Malith-19, line 154: Check we are iterating all of the fields when getting

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
Malith-19: check if this is actual application id or a different one

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DeviceTokenService.java
Malith-19: Check if we can pass the public key as a parameter to same evaluate method and without having this class.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/resolver/DeviceDataResolverImpl.java
Malith-19: Since we are recommending the header for the token, we should be checking the header first and then as a fallback we should be checking parameter.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
Malith-19: Check we can handle the "OR" scenario inside the missingField check logic or policy evaluation logic.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/js/DevicePolicyJsFunction.java
Malith-19: Fixed with https://github.com/wso2/carbon-identity-framework/pull/8228/commits/4695f356a9d1e42056f0e7f35ed27c05e009e9c6

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
Malith-19: Fixed by https://github.com/wso2/carbon-identity-framework/pull/8228/commits/7087653f1d27f1d6c3846c955487a2898a43da03

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/util/DeviceTokenExtractor.java
Malith-19: The device lookup can't leave this class, since device.getPublicKey() is what the token's ECDSA signature is verified against. Given the record is already loaded here, the status check is two lines on it.

Moving just the status check to evaluate(...) isn't straightforward either: the deviceId comes from the JWT header (getHeader().getCustomParam("deviceId")), and only the claims are copied into the returned device data map. So evaluate has no device identity to check — it receives (policyId, deviceData, appId, tenantDomain), and deviceId/status aren't DeviceFields either. We'd have to plumb the id (or the Device) out of the extractor and do a second lookup.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DevicePolicyEvaluator.java
Malith-19: Checked and this is application id.

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/api/service/DeviceTokenService.java
Malith-19: Moved by https://github.com/wso2/carbon-identity-framework/pull/8228/commits/1ad31eeed734c97175257666a13e9b2cebfc2c61

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/resolver/DeviceDataResolverImpl.java
Malith-19: Fixed by https://github.com/wso2/carbon-identity-framework/pull/8228/commits/de48846fb9d3c535b48e5fc9713fd7a179b53c86

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
Malith-19: Fixed with https://github.com/wso2/carbon-identity-framework/pull/8228/commits/001ad743da9d8138d6a3ffa60ee38b69de15020b

components/device-mgt/org.wso2.carbon.identity.device.policy/src/main/java/org/wso2/carbon/identity/device/policy/internal/service/impl/DevicePolicyEvaluatorImpl.java
Malith-19: Fixed by https://github.com/wso2/carbon-identity-framework/pull/8228/commits/39b254fd20d08ad8a121554d0b4c2a223e7b8d9d

Approach: moved the determination after evaluation. Only the platform is now required up front, since it picks the rule; everything else goes to the engine. If the result comes back unsatisfied, the failed fields that the device never sent are reported as INCOMPLETE_DEVICE_DATA, and the rest as NON_COMPLIANT. That way a policy that can pass on partial data passes, and we still distinguish "you didn't send this" from "this breached the policy".

Thisara-Welmilla (approved):

PR #8259, opened 2026-08-20
Remove the multiple client secrets enable toggle so the feature is always available

AfraHussaindeen:
### Purpose

Follow-up to #8226. Removes the server-level `Enable` toggle for the multiple client secrets feature so the feature is always available. The `MaxSecretCount` capacity ceiling is retained.

### Changes

- **`identity.xml.j2`** — drop `<Enable>` from the `<MultipleClientSecrets>` block (keep `<MaxSecretCount>`) and update the accompanying config comment.
- **`default.json`** — remove `oauth.multiple_client_secrets.enable` (keep `oauth.multiple_client_secrets.max_secret_count`).
- **`infer.json`** — remove the `oauth.multiple_client_secrets.enable = false` overrides for `IS_7.0.0`–`IS_7.3.0`.

### Related PR

- #8226

### Related Issue 
 - https://github.com/wso2/product-is/issues/28127

### DIFF
--- a/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
+++ b/features/identity-core/org.wso2.carbon.identity.core.server.feature/resources/identity.xml.j2
@@ -298,13 +298,13 @@
         {% else %}
         <RecomputeSubjectClaimForAlternateSubjectIdentifier>false</RecomputeSubjectClaimForAlternateSubjectIdentifier>
         {% endif %}
-        <!-- Multi-secret (multiple client secrets per OAuth application) config.
-             `Enable` toggles the feature at the server level.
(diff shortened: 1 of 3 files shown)


sadilchamishka (approved):

pavinduLakshan:
When a new feature is introduced, wouldn't it be better to keep this config and set it's default value to true? Then we can conditionally turn it off when required. WDYT?

AfraHussaindeen:
> When a new feature is introduced, wouldn't it be better to keep this config and set it's default value to true? Then we can conditionally turn it off when required. WDYT?

When the secret limit is set to 1, the single secret behaviour in the UI will be preserved.

PR #8266, opened 2026-08-27
Resolve tenant owner from the tenant store in Administrator role guards

sadilchamishka:
## Issue

The guards that enforce "only the tenant owner can remove users/groups from the Administrator role"
compare the caller against `RealmConfiguration.getAdminUserName()`, which is served from the
node-local realm cache. That cache is populated when the tenant realm is built and is never refreshed
when the tenant's admin user changes underneath a running node. Once that happens the decision
inverts: the current owner gets `403 "Invalid operation. Only the tenant owner can remove users from
the role: Administrator"`, while the previous owner keeps full owner authority — until the realm is
evicted or the node is restarted.

## Fix

Inside the administrator-role branch of `RoleManagementServiceImpl.validateUserRemovalFromRole` and
`RoleDAOImpl.validateGroupRemovalFromRole`, re-resolve the owner from the tenant store via a new
`RoleManagementUtils.resolveTenantOwnerUsername(...)` instead of trusting the cached realm
configuration. It reads a single tenant row and leaves the realm itself alone, and it is only reached
for the administrator role. If the tenant store cannot answer — tenant absent, no realm config, or a
`UserStoreException` — the cached value is returned unchanged, so nothing else changes behaviour. The
super tenant short-circuits, since its realm comes from the bootstrap configuration rather than
`UM_TENANT`.

## Testing

8 new unit cases covering the stale, current, super-tenant, and every fallback path (module suite
63 → 71, all green). Also verified end-to-end on a local IS 7.4.0-SNAPSHOT: after changing the
tenant's admin user with the node running, the new owner goes 403 → 200 and the previous owner
200 → 403, with no restart.

### DIFF
--- a/components/role-mgt/org.wso2.carbon.identity.role.v2.mgt.core/src/main/java/org/wso2/carbon/identity/role/v2/mgt/core/RoleManagementServiceImpl.java
+++ b/components/role-mgt/org.wso2.carbon.identity.role.v2.mgt.core/src/main/java/org/wso2/carbon/identity/role/v2/mgt/core/RoleManagementServiceImpl.java
@@ -1179,9 +1179,17 @@ public class RoleManagementServiceImpl implements RoleManagementService {
                     && RoleConstants.ORGANIZATION.equals(role.getAudience()))
                     || (RoleConstants.ADMINISTRATOR.equals(role.getName()) &&
                             RoleConstants.CONSOLE_APP_AUDIENCE_NAME.equals(role.getAudienceName()))) {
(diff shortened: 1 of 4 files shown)


sadilchamishka:
Closing this. The ownership transfer that triggers the reported behaviour is a database-only operation (a support-run script), so no product code executes at transfer time and there is no hook from which to invalidate the tenant realm cache. The documented transfer procedure already includes a rolling restart, which is the cache-clearing step. Hardening the role guards to re-read the owner on every administrator-role change is therefore not the right trade-off for this case.

The branch remains on the fork if this is revisited.

PR #8277, opened 2026-09-06
Fix application certificate deletion/update not being transactional

### LINKED ISSUE #28238: Removing an application certificate is not transactional: a failed application update leaves a dangling `SP_METADATA` `CERTIFICATE` reference that permanently breaks the application and the whole application list
### Description

Removing an application certificate performs **two writes in two different transactions**. The certificate record is deleted on its own connection and committed immediately, while the removal of the application's reference to it is only persisted at the very end of the application update, on the shared transactional connection.

If **anything** fails in between, the application update rolls back — but the certificate deletion cannot be rolled back. The application is left referencing a certificate record that no longer exists.

From that point on the application is **permanently unusable and unrepairable through the management APIs**:

- `GET /api/server/v1/applications?attributes=advancedConfigurations` → **`HTTP 500`**, failing the **entire list**, not just the affected application
- `GET /api/server/v1/applications/{id}` → `HTTP 500`
- `PATCH /api/server/v1/applications/{id}` → `HTTP 500`, **including a request that clears the certificate**, and including requests that do not touch the certificate at all
- `DELETE /api/server/v1/applications/{id}` → `HTTP 500`, so the application cannot even be removed and recreated

⬦ The Console application list is therefore unusable for the whole tenant until the affected application is repaired directly in the database, which is the only available recovery.

### Root cause

An application certificate is not stored on `SP_APP`. It is a row in `SP_METADATA` with `NAME='CERTIFICATE'` whose `VALUE` is the `IDN_CERTIFICATE.ID`:

```java
private static final String SP_PROPERTY_NAME_CERTIFICATE = "CERTIFICATE";
```

`updateApplicationCertificate` routes a blank certificate to `removeCertificateReferenceAndDelete`, which mutates the **in-memory** property array and deletes the certificate record:

```java
private void updateApplicationCertificate(ServiceProvider serviceProvider, int tenantID) {
    if (StringUtils.isBlank(serviceProvider.getCertificateContent())) {
        // Remove the certificate reference property if exists and remove the certificate.
        removeCertificateReferenceAndDelete(serviceProvider, tenantID);
    } else { ... }
}

private void removeCertificateReferenceAndDelete(ServiceProvider serviceProvider, int tenantID) {
    ...
    serviceProvider.setSpProperties(getFilteredSpProperties(spProperties, certificateReferenceIndex));  // in memory only
    deleteCertificate(certificateID, IdentityTenantUtil.getTenantDomain(tenantID));                     // committed immediately
}
```

The delete goes through `ApplicationCertificateManagementService`, whose DAO opens its **own** connection from the datasource and commits its **own** transaction — it does not join the caller's transaction:

```java
public void deleteCertificate(int certificateId, int tenantId) throws CertificateMgtException {
    NamedJdbcTemplate jdbcTemplate = new NamedJdbcTemplate(IdentityDatabaseUtil.getDataSource());
    jdbcTemplate.withTransaction(template -> { template.executeUpdate(DELETE_CERTIFICATE, ...); return null; });
}
```

Within `addApplicationConfigurations` the certificate delete happens **early**, and the property array is written to `SP_METADATA` **last**:

```java
updateBasicApplicationData(serviceProvider, connection);
updateApplicationCertificate(serviceProvider, tenantID);          // ← certificate row deleted + committed here
updateInboundProvisioningConfiguration(...);
updateInboundAuthRequestConfiguration(...);
updateLocalAndOutboundAuthenticationConfiguration(...);
updateRequestPathAuthenticators(...);
updateClaimConfiguration(...);
updateOutboundProvisioningConfiguration(...);
updateSpTrustedAppMetadata(...);
... permissions / roles / associated roles ...
updateConfigurationsAsServiceProperties(serviceProvider);
if (ArrayUtils.isNotEmpty(serviceProvider.getSpProperties())) {   // ← guard, see "Related defects"
    updateServiceProviderProperties(connection, applicationId, Arrays.asList(spProperties), tenantID);
}
```

Every step between those two points can fail (constraint violation, lock timeout, deadlock, DB failover, listener error). When one does, the transaction on `connection` rolls back and `SP_METADATA` keeps `CERTIFICATE=<id>`, while `IDN_CERTIFICATE` no longer has that row.

Reads then fail because `getCertificateContent` finds the reference, gets `null` back, and the certificate service raises a client exception:

```java
public Certificate getCertificate(int certificateId, String tenantDomain) throws CertificateMgtException {
    Certificate certificate = CACHE_BACKED_DAO.getCertificate(certificateId, IdentityTenantUtil.getTenantId(tenantDomain));
    if (certificate == null) {
        CertificateMgtExceptionHandler.throwClientException(CertificateMgtErrors.ERROR_CERTIFICATE_DOES_NOT_EXIST,
                String.valueOf(certificateId));
    }
    return certificate;
}
```

`ERROR_CERTIFICATE_DOES_NOT_EXIST` is `("60001", "Unable to perform the operation.", "Certificate with the id: %s does not exist.")`, which is why the surfaced message is the unhelpfully generic *"Unable to perform the operation."*

**Blast radius.** `getSpWithRequiredAttributes` iterates every application in the page with no per-application error handling, so one unreadable application fails the whole response:

```java
for (ApplicationBasicInfo applicationBasicInfo : filteredAppList) {
    ServiceProvider serviceProvider = applicationManagementService
            .getApplicationWithRequiredAttributes(applicationBasicInfo.getApplicationId(), requestedAttributeList);
    ...
}
```

### Affected code

- `org.wso2.carbon.identity.application.mgt.dao.impl.ApplicationDAOImpl#updateApplicationCertificate` / `#removeCertificateReferenceAndDelete` — deletes the certificate record outside the application-update transaction.
- `ApplicationDAOImpl#addApplicationConfigurations` — certificate delete near the start, `updateServiceProviderProperties` at the end, with many failure points in between.
- `org.wso2.carbon.identity.certificate.management.dao.impl.ApplicationCertificateManagementDAOImpl#deleteCertificate` (and `CertificateManagementDAOImpl#deleteCertificate`) — `new NamedJdbcTemplate(IdentityDatabaseUtil.getDataSource())` + `withTransaction`, i.e. an independent transaction.
- `ApplicationDAOImpl#getCertificateContent` — wraps the missing certificate as `CertificateRetrievingException`, failing the whole application read.
- `org.wso2.carbon.identity.api.server.application.management.v1.core.ServerApplicationManagementService#getSpWithRequiredAttributes` — no per-application isolation.

### Stacktrace

⬦ Observed on an affected 7.1.0 deployment (application `id: 7`). The `Caused by` chain is the invariant part; the specific failure that created the dangling reference is not visible here.

```
ERROR {org.wso2.carbon.identity.api.server.application.management.v1.core.functions.Utils} - errorCode: APP-65001 | message: Failed to gather required attributes for application with id: 7
org.wso2.carbon.identity.application.common.IdentityApplicationManagementException: Failed to gather required attributes for application with id: 7
    at org.wso2.carbon.identity.application.mgt.dao.impl.ApplicationDAOImpl.getApplicationWithRequiredAttributes(ApplicationDAOImpl.java:2478)
    at org.wso2.carbon.identity.application.mgt.ApplicationManagementServiceImpl.getApplicationWithRequiredAttributes(ApplicationManagementServiceImpl.java:1334)
    at org.wso2.carbon.identity.api.server.application.management.v1.core.ServerApplicationManagementService.getSpWithRequiredAttributes(ServerApplicationManagementService.java:406)
    at org.wso2.carbon.identity.api.server.application.management.v1.core.ServerApplicationManagementService.getAllApplications(ServerApplicationManagementService.java:327)
    at org.wso2.carbon.identity.api.server.application.management.v1.impl.ApplicationsApiServiceImpl.getAllApplications(ApplicationsApiServiceImpl.java:96)
    at org.wso2.carbon.identity.api.server.application.management.v1.ApplicationsApi.getAllApplications(ApplicationsApi.java:518)
    ...
Caused by: org.wso2.carbon.identity.core.CertificateRetrievingException: An error occurred while retrieving the certificate for the application.
    at org.wso2.carbon.identity.application.mgt.dao.impl.ApplicationDAOImpl.getCertificateContent(ApplicationDAOImpl.java:1943)
    at org.wso2.carbon.identity.application.mgt.dao.impl.ApplicationDAOImpl.getApplicationWithRequiredAttributes(ApplicationDAOImpl.java:2443)
    ... 63 more
Caused by: org.wso2.carbon.identity.certificate.management.exception.CertificateMgtClientException: Unable to perform the operation.
    at org.wso2.carbon.identity.certificate.management.util.CertificateMgtExceptionHandler.throwClientException(CertificateMgtExceptionHandler.java:50)
    at org.wso2.carbon.identity.certificate.management.service.impl.ApplicationCertificateManagementServiceImpl.getCertificate(ApplicationCertificateManagementServiceImpl.java:95)
    at org.wso2.carbon.identity.application.mgt.dao.impl.ApplicationDAOImpl.getCertificateContent(ApplicationDAOImpl.java:1934)
    ... 64 more
```

### Steps to Reproduce

**Any** failure occurring after the certificate delete reproduces this. The steps below use a JWKS URL longer than `SP_METADATA.VALUE VARCHAR(255)` purely because it is a deterministic, single-call way to force such a failure — the URL length is not the subject of this report.

The switch from PEM to JWKS is what triggers the certificate deletion, since the JWKS branch nulls the certificate content:

```java
} else if (TYPE_JWKS.equals(certificate.getType())) {
    setIfNotNull(certificate.getValue(), serviceProvider::setJwksUri);
    serviceProvider.setCertificateContent(null);        // → StringUtils.isBlank(...) → certificate record deleted
}
```

…and the JWKS URI is itself persisted into `SP_METADATA.VALUE`, which is `VARCHAR(255)`:

```java
jwksUri.setValue(StringUtils.isNotBlank(sp.getJwksUri()) ? sp.getJwksUri() : StringUtils.EMPTY);
```

1. WSO2 Identity Server 7.1.0 (reproduced on update level 77), default configuration.

2. Create an application with a PEM certificate:

   ```bash
   curl -k -u admin:admin -X POST 'https://localhost:9443/api/server/v1/applications' \
     -H 'Content-Type: application/json' \
     --data '{"name":"cert-orphan-repro","templateId":"custom-application-oidc",
              "advancedConfigurations":{"certificate":{"type":"PEM","value":"-----BEGIN CERTIFICATE-----\n...\n-----END CERTIFICATE-----\n"}},
              "inboundProtocolConfiguration":{"oidc":{"grantTypes":["client_credentials"],"publicClient":false}}}'
   ```

   Confirm the pair exists — `IDN_CERTIFICATE` has one row, and `SP_METADATA` holds the matching reference:

   ```sql
   SELECT ID, NAME FROM IDN_CERTIFICATE;
   SELECT SP_ID, NAME, "VALUE" FROM SP_METADATA WHERE NAME = 'CERTIFICATE';
   ```

3. Switch the certificate from PEM to a JWKS URL of 298 characters (equivalent to *Application → Certificate → Use JWKS endpoint* in the Console):

   ```bash
   curl -k -u admin:admin -X PATCH 'https://localhost:9443/api/server/v1/applications/{id}' \
     -H 'Content-Type: application/json' \
     --data '{"advancedConfigurations":{"certificate":{"type":"JWKS","value":"https://idp.example.com/oauth2/jwks?p=aaaa…(298 chars total)"}}}'
   ```

   The request returns **`HTTP 500`** (`APP-65001`), and `wso2carbon.log` shows the failure that rolls the update back:

   ```
   Caused by: org.h2.jdbc.JdbcBatchUpdateException: Value too long for column "VALUE CHARACTER VARYING(255)":
   "'https://idp.example.com/oauth2/jwks?p=aaaaaaaa... (298)"; SQL statement: ...
   ```

4. **Observe the database is now inconsistent, with no manual modification having been made:**

   ```sql
   SELECT COUNT(*) FROM IDN_CERTIFICATE;                             -- 0  (deleted and committed)
   SELECT SP_ID, "VALUE" FROM SP_METADATA WHERE NAME = 'CERTIFICATE'; -- row still present (rolled back)
   ```

5. **Observe the application is now permanently broken.** With `{n}` as the affected `SP_APP.ID`:

   | Request | Result |
   |---|---|
   | `GET /api/server/v1/applications?attributes=advancedConfigurations` | `500` — *Failed to gather required attributes for application with id: {n}*, **entire list fails** |
   | `GET /api/server/v1/applications` (no `attributes`) | `200` |
   | `GET /api/server/v1/applications/{id}` | `500` |
   | `PATCH /api/server/v1/applications/{id}` clearing the certificate | `500` |
   | `PATCH /api/server/v1/applications/{id}` with an unrelated body (e.g. `{"description":"x"}`) | `500` |
   | `DELETE /api/server/v1/applications/{id}` | `500` |
   | Repeating step 3 with a short URL | `500` — **cannot self-heal** |

6. The only recovery is a direct database correction, after which every call above succeeds immediately with no restart:

   ```sql
   DELETE FROM SP_METADATA WHERE SP_ID = {n} AND NAME = 'CERTIFICATE';
   ```

**Control (baseline).** Performing the same certificate removal **without** any induced failure (`{"advancedConfigurations":{"certificate":{"type":"PEM","value":""}}}`) returns `200`, removes **both** the certificate record and the reference, and leaves the application list healthy. The inconsistency is specific to a failure occurring after the certificate delete.

**Second, independent trigger (same end state).** Instead of the oversized URL, holding an uncommitted write lock on the application's `SP_METADATA` rows from a separate session so the final `updateServiceProviderProperties` hits `LOCK_TIMEOUT` produces an identical outcome: `IDN_CERTIFICATE` row gone, reference retained. Two unrelated failures, one defect.

### Expected Behavior

Removing an application certificate should be atomic with the application update. Either both the certificate record and its reference are removed, or neither is.

A failed application update must never leave the application in a state where it cannot be read, updated, or deleted through the management APIs. Additionally, one unreadable application should not cause `GET /api/server/v1/applications` to fail for every application in the tenant.

### Suggested Fix

1. **Make the certificate removal participate in the application-update transaction.** Either pass the caller's `Connection` down to the certificate management DAO, or defer the certificate delete until after the application-update transaction commits, or compensate (re-create the certificate record) if the update rolls back. Deferring is the least invasive: collect the certificate id to delete during `addApplicationConfigurations` and perform the delete only after `IdentityDatabaseUtil.commitTransaction(connection)` succeeds.

2. **Isolate per-application failures in the listing.** In `getSpWithRequiredAttributes`, catch per application so one broken application degrades that entry rather than failing the whole page.

These have not been validated as patches.

### Related defects observed

Both were found while investigating the above and may warrant separate issues:

- **No length validation on the JWKS URI.** The API accepts a JWKS URL it cannot persist, since `SP_METADATA.VALUE` is `VARCHAR(255)`. The request fails with an opaque `HTTP 500` (`APP-65001`) and a raw JDBC truncation error rather than a `400`. This is the trigger used in the steps above.
- **The SP-properties write is conditionally skipped.** `if (ArrayUtils.isNotEmpty(serviceProvider.getSpProperties()))` guards `updateServiceProviderProperties`. If the property array is empty after the certificate property is removed, the `SP_METADATA` rewrite is skipped entirely and the stale reference survives *with no failure at all*. In practice applications carry other properties (`updateConfigurationsAsServiceProperties` runs immediately before), so this path is unlikely to be reached — but it is the same defect class and the guard looks unsafe.

### Please select the area issue is related to

Identity Server Core

### Version

IS 7.1.0 (potentially 7.2.0 and 7.3.0)

### Environment Details (with versions)

- WSO2 Identity Server 7.1.0, update level 77, single node
- JDK 11 (Temurin 11.0.31)
- ⬦ H2 (bundled) for the reproduction; the defect is database-independent — the DB only determines *which* failures can occur between the certificate delete and the property write. It has also been observed on a different RDBMS.

### Developer Checklist

- [ ] [Behavioural Change] Does this change introduce a behavioral change to the product?
- [ ]  ↳ Approved by team lead
- [ ]  ↳ Label `impact/behavioral-change` added
- [ ]  ↳ Migration label added (e.g., `7.2.0-migration`)
- [ ]  ↳ Migration issues created and linked
- [ ]  ↳ Label `config` added
- [ ]  ↳ Configuration is properly documented

pavinduLakshan:
## Summary
- Application certificate delete/update ran on its own, independently committed connection instead of the enclosing application transaction's connection.
- If a later step in that transaction failed and rolled back, the `SP_METADATA` `CERTIFICATE` reference survived while the `IDN_CERTIFICATE` row it pointed to was already deleted/updated, permanently breaking reads/updates/deletes of the affected application.
- Threads the caller's `Connection` through `ApplicationCertificateManagementService`/DAO so certificate writes commit and roll back together with the application transaction.

Fixes wso2/product-is#28238

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,15 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+    {
+        "type": "java",
+        "name": "Java Remote Debugger",
+        "request": "attach",
+        "hostName": "127.0.0.1",
+        "port": "5005"
+    }
+    ]
+}
--- a/components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/dao/impl/ApplicationDAOImpl.java
+++ b/components/application-mgt/org.wso2.carbon.identity.application.mgt/src/main/java/org/wso2/carbon/identity/application/mgt/dao/impl/ApplicationDAOImpl.java
@@ -629,7 +629,7 @@ public class ApplicationDAOImpl extends AbstractApplicationDAOImpl implements Pa
         // you can change application name, description, isSasApp...
         updateBasicApplicationData(serviceProvider, connection);
 
-        updateApplicationCertificate(serviceProvider, tenantID);
+        updateApplicationCertificate(serviceProvider, tenantID, connection);
 
         updateInboundProvisioningConfiguration(applicationId, serviceProvider.getInboundProvisioningConfig(),
                 connection);
@@ -670,10 +670,10 @@ public class ApplicationDAOImpl extends AbstractApplicationDAOImpl implements Pa
         }
 
         updateConfigurationsAsServiceProperties(serviceProvider);
-        if (ArrayUtils.isNotEmpty(serviceProvider.getSpProperties())) {
-            ServiceProviderProperty[] spProperties = serviceProvider.getSpProperties();
-            updateServiceProviderProperties(connection, applicationId, Arrays.asList(spProperties), tenantID);
-        }
+        ServiceProviderProperty[] spPropertiesArr = serviceProvider.getSpProperties();
+        List<ServiceProviderProperty> spProperties = spPropertiesArr == null
+                ? Collections.emptyList() : Arrays.asList(spPropertiesArr);
+        updateServiceProviderProperties(connection, applicationId, spProperties, tenantID);
 
         // Will be supported with 'Advance Consent Management Feature'.
             /*
@@ -798,19 +798,21 @@ public class ApplicationDAOImpl extends AbstractApplicationDAOImpl implements Pa
      *
      * @param serviceProvider Service provider object.
      * @param tenantID        Tenant ID.
+     * @param connection      The connection used for the enclosing application-update transaction.
      * @throws IdentityApplicationManagementException If an error occurs while updating the certificate.
      */
-    private void updateApplicationCertificate(ServiceProvider serviceProvider, int tenantID)
+    private void updateApplicationCertificate(ServiceProvider serviceProvider, int tenantID, Connection connection)
             throws IdentityApplicationManagementException {
 
         if (StringUtils.isBlank(serviceProvider.getCertificateContent())) {
             // Remove the certificate reference property if exists and remove the certificate.
-            removeCertificateReferenceAndDelete(serviceProvider, tenantID);
+            removeCertificateReferenceAndDelete(serviceProvider, tenantID, connection);
         } else {
             String certificateReferenceIdString = getCertificateReferenceID(serviceProvider.getSpProperties());
             if (certificateReferenceIdString != null) {
                 // If there is a reference, update the relevant existing certificate record.
-                updateCertificate(certificateReferenceIdString, serviceProvider.getCertificateContent(), tenantID);
+                updateCertificate(certificateReferenceIdString, serviceProvider.getCertificateContent(), tenantID,
+                        connection);
             } else {
                 // There is no existing reference. Persisting the certificate as a new record.
                 persistApplicationCertificate(serviceProvider, tenantID);
@@ -820,13 +822,15 @@ public class ApplicationDAOImpl extends AbstractApplicationDAOImpl implements Pa
 
     /**
      * Removes the certificate reference property from the given service provider object and deletes the certificate
-     * record from the database.
+     * record from the database, on the given connection so the delete becomes part of the caller's transaction.
      *
      * @param serviceProvider Service provider object.
      * @param tenantID        Tenant ID.
+     * @param connection      Connection of the enclosing application-update transaction.
      * @throws IdentityApplicationManagementException If an error occurs while removing the certificate reference.
      */
-    private void removeCertificateReferenceAndDelete(ServiceProvider serviceProvider, int tenantID)
+    private void removeCertificateReferenceAndDelete(ServiceProvider serviceProvider, int tenantID,
+                                                      Connection connection)
             throws IdentityApplicationManagementException {
 
         ServiceProviderProperty[] spProperties = serviceProvider.getSpProperties();
@@ -844,7 +848,7 @@ public class ApplicationDAOImpl extends AbstractApplicationDAOImpl implements Pa
             int certificateID = Integer.parseInt(spProperties[certificateReferenceIndex].getValue());
 
             serviceProvider.setSpProperties(getFilteredSpProperties(spProperties, certificateReferenceIndex));
-            deleteCertificate(certificateID, IdentityTenantUtil.getTenantDomain(tenantID));
+            deleteCertificate(certificateID, IdentityTenantUtil.getTenantDomain(tenantID), connection);
         }
     }
 
@@ -868,20 +872,23 @@ public class ApplicationDAOImpl extends AbstractApplicationDAOImpl implements Pa
     }
 
     /**
-     * Update the existing certificate record with the given certificate ID.
+     * Update the existing certificate record with the given certificate ID, on the given connection so the update
+     * becomes part of the caller's transaction.
      *
      * @param certificateId      Certificate ID.
      * @param certificateContent Certificate content to be updated.
      * @param tenantID           Tenant ID.
+     * @param connection         Connection of the enclosing application-update transaction.
      * @throws IdentityApplicationManagementException If an error occurs while updating the certificate.
      */
-    private void updateCertificate(String certificateId, String certificateContent, int tenantID)
+    private void updateCertificate(String certificateId, String certificateContent, int tenantID,
+                                   Connection connection)
             throws IdentityApplicationManagementException {
 
         try {
             ApplicationManagementServiceComponentHolder.getInstance().getApplicationCertificateMgtService()
                     .updateCertificateContent(Integer.parseInt(certificateId), certificateContent,
-                            IdentityTenantUtil.getTenantDomain(tenantID));
+                            IdentityTenantUtil.getTenantDomain(tenantID), connection);
         } catch (CertificateMgtClientException e) {
             throw new IdentityApplicationManagementClientException(INVALID_REQUEST.getCode(), e.getDescription(), e);
         } catch (CertificateMgtException e) {
@@ -4345,7 +4352,7 @@ public class ApplicationDAOImpl extends AbstractApplicationDAOImpl implements Pa
         try {
 
             // Delete the application certificate if there is any.
-            deleteCertificate(appName, tenantID);
+            deleteCertificate(appName, tenantID, connection);
 
             // First, delete all the clients of the application
             int applicationID = getApplicationIDByName(appName, tenantID, connection);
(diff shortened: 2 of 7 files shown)


pavinduLakshan:
Closing as passing db connections between separate OSGI services (app mgt and cert mgt) is an anti pattern

PR #8286, opened 2026-09-15
Fix invalid BIGINT datatype in Oracle RAC identity script

ManadaHerath:
## Purpose

Fixes the Oracle RAC identity DB script failing with `ORA-00902: invalid datatype`.

`IDN_OAUTH2_REFRESH_TOKEN.REFRESH_TOKEN_VALIDITY_PERIOD` was declared `BIGINT`, which is not a
valid Oracle datatype. The `CREATE TABLE` fails and 8 dependent objects then fail with
`ORA-00942`, so the refresh-token schema is never created on a fresh Oracle install.

Also removes a stray `END; /` block after `IDN_VC_CLAIMS_TRIG` (the trigger already closes
correctly; the orphan produced a spurious `SP2-0042`).

Public: wso2/product-is#28358

## Changes

`resources/dbscripts/oracle_rac.sql`

```diff
-    REFRESH_TOKEN_VALIDITY_PERIOD BIGINT,
+    REFRESH_TOKEN_VALIDITY_PERIOD NUMBER(19),

     END;
 /
-    END;
-/

```


RushanNanayakkara (approved):

RushanNanayakkara:
Merging despite the SQL Table Monitor faiilure since this check has been failing due to an issue in the action itself for some time.

PR #8287, opened 2026-09-15
Add token exchange certificate expiry validity configuration

DilshanSenarath:
### Purpose

Adds a token exchange specific configuration that controls whether certificate expiry is enforced when a trusted token issuer is configured with a manually added certificate.

`TokenExchange.EnforceCertificateExpiryValidity` applies only to the certificate based signature validation path. JWKS based validation is not affected and continues to use its own configuration.

Default is `true`, so behaviour is unchanged unless it is explicitly disabled:

```toml
[token_exchange]
enforce_certificate_expiry_validity = false
```

### Changes

- `identity.xml.j2` - new element in the existing `<TokenExchange>` block
- `default.json` - default value
- `identity.xml` - reference sample of the generated output

Required by https://github.com/wso2/product-is/issues/28360


DilshanSenarath:
Closing this.

The agreed approach is to honour the existing `JWTValidatorConfigs.EnforceCertificateExpiryTimeValidity` configuration in token exchange, rather than introducing a token exchange specific one. That configuration is already honoured on the manually configured certificate path in OIDC federated authentication, federated IdP-initiated logout and the JWT bearer grant, so no new configuration is needed and this change is unnecessary.

PR #8297, opened 2026-09-22
Read the sessions of a session ID list in batches

sadilchamishka:

Listing the sessions of a user resolved each session ID on its own, and each of those reads queried the session application info, the session metadata and the application names separately. The cost of a listing therefore grew with the number of sessions the account held, which matters on the login path, because a listing happens there whenever an application's adaptive script performs a concurrent session check.

- `UserSessionDAOImpl.getSessions(List<String>)` reads the application info and the metadata of all the given session IDs in batches, and resolves the application names once for the whole set rather than once per session. Batch sizes stay below the parameter limits of the supported databases.
- `UserSessionManagementServiceImpl.getActiveSessionList` reads the sessions up front and looks each one up while iterating.

The number of queries a listing issues no longer depends on how many sessions the account holds.

Behaviour is unchanged: the same session ID query, the same session context lookups, the same effective-session filtering, and the same rule that a session is not returned unless it has an application that exists.

The single session reads delegate to the batched read, so the session application info lookup and the session property mapping each have one implementation rather than three. 

### Follow up actions

- The session context of each session is still read individually, which is a cache lookup that falls back to reading the session object on a miss. Batching that is a separate change.

### Developer Checklist (Mandatory)

- [x] Unit tests added for the batched read: a list spanning more than one batch, sessions without a
      resolvable application being skipped, and blank or empty input.
- [x] No configuration, schema or migration change.
- [x] No change to any public API contract; the new DAO method is a default interface method.


components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
```diff
@@ -661,13 +661,13 @@ private List<UserSession> getActiveSessionList(List<String> sessionIdList, Strin
             throws SessionManagementServerException {
 
         List<UserSession> sessionsList = new ArrayList<>();
+        Map<String, UserSession> userSessions = new UserSessionDAOImpl().getSessions(sessionIdList);
         for (String sessionId : sessionIdList) {
```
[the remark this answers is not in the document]
sadilchamishka: Fixed in 57ca804. Blank and null session IDs are now dropped before the batch is built, so they cannot take a parameter of the IN clause, and an all-blank list returns without querying. Covered by `testGetSessionsWithBlankAndEmptyInput`.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/dao/impl/UserSessionDAOImpl.java
```diff
@@ -59,55 +60,67 @@ public class UserSessionDAOImpl implements UserSessionDAO {
...
+    private static final int SESSION_ID_BATCH_SIZE = 500;
+
     public UserSessionDAOImpl() {
     }
 
+    /**
+     * Retrieve the sessions of the given session IDs, reading the session application info and the session metadata
+     * of all of them in batches instead of a query per session ID.
+     *
+     * @param sessionIds Session IDs.
```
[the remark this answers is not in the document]
sadilchamishka: Added in 57ca804, in `UserSessionDAOImplTest`: a list of 501 session IDs spanning more than one batch and asserting sessions from the first and the last batch come back; a session with no application info and a session whose only application is not in SP_APP both being skipped; and blank or empty input. The second of those catches a real defect this PR had introduced, which is described in the reply on the merge risk comment.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/dao/impl/UserSessionDAOImpl.java
```diff
@@ -59,55 +60,67 @@ public class UserSessionDAOImpl implements UserSessionDAO {
...
                     SessionMgtConstants.ErrorMessages.ERROR_CODE_UNABLE_TO_GET_SESSION,
                     SessionMgtConstants.ErrorMessages.ERROR_CODE_UNABLE_TO_GET_SESSION.getDescription(), e);
         }
-        return null;
+
+        applicationsBySession.forEach((sessionId, applications) -> {
+            /*
+            A session without application info is not returned, which is what a single session read does when its
+            application list is empty.
+             */
```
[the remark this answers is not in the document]
sadilchamishka: Leaving this one as it is. The sentence does end with a period; it is wrapped across two lines because it exceeds the line length, which is the existing style for block comments in this component (for example the `FilterByUniqueSessionIdForUser` comment in `UserSessionManagementServiceImpl`). Happy to reflow it if a reviewer prefers otherwise.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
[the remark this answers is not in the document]
sadilchamishka: Correcting my earlier reply on this thread: the blank filtering is removed again in 8ad4d94, because it was guarding against something that cannot happen and would be harmless if it did.

IDN_AUTH_USER_SESSION_MAPPING has a unique constraint on (USER_ID, SESSION_ID) and every caller reads the IDs of a single user, so the list is already unique, and SESSION_ID is not nullable. If a null or a duplicate did arrive, IN matches a row once and the result is keyed by session ID, so a duplicate costs one parameter and a null matches nothing, leaving that session out of the result, which is what the caller already handles with its own null check.

`testGetSessionsWithBlankAndEmptyInput` still passes unchanged, since it asserts the behaviour rather than the filtering.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/dao/impl/UserSessionDAOImpl.java
```diff
@@ -59,55 +59,75 @@ public class UserSessionDAOImpl implements UserSessionDAO {
...
-                    case SessionMgtConstants.LOGIN_TIME:
-                        userSession.setLoginTime(value);
-                        break;
-                }
-            });
-
-            if (!applicationList.isEmpty()) {
-                userSession.setApplications(applicationList);
-                return userSession;
+            for (int start = 0; start < sessionIds.size(); start += SESSION_ID_BATCH_SIZE) {
```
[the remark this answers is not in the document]
sadilchamishka: Added in the latest commit. The class had no logger, so it now has one following the convention of the sibling store class, and the batch loop logs the batch size at DEBUG before the two reads. No session IDs are logged.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/dao/impl/UserSessionDAOImpl.java
```diff
@@ -354,13 +365,38 @@ private void generateApplicationFromAppID(List<Application> applications) throws
...
                 applications.stream().collect(Collectors.groupingBy(Application::getAppId));
-        String placeholder = String.join(", ", Collections.nCopies(appIdMap.keySet().size(), "?"));
+        List<String> appIds = new ArrayList<>(appIdMap.keySet());
+        /*
+         * Read in batches: this is called with the applications of every session of a user, so the number of distinct
+         * applications is not bounded by what one session has.
+         */
+        for (int start = 0; start < appIds.size(); start += APPLICATION_ID_BATCH_SIZE) {
+            List<String> batch = appIds.subList(start, Math.min(start + APPLICATION_ID_BATCH_SIZE, appIds.size()));
+            readApplicationNames(batch, appIdMap);
```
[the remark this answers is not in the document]
sadilchamishka: This one no longer applies: `readApplicationNames` has been removed. Batching the application lookup was reverted, because `generateApplicationFromAppID` is unrelated to reading sessions in batches and modifying it widened this change unnecessarily. That method is now byte identical to master again.

The application names are instead resolved once per batch of sessions inside the batch loop, so the untouched method keeps receiving a list no larger than one batch can produce, and the DEBUG log added for the batch covers that read as well.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/internal/impl/UserSessionManagementServiceImpl.java
```diff
@@ -661,13 +661,13 @@ private List<UserSession> getActiveSessionList(List<String> sessionIdList, Strin
...
             if (sessionId != null) {
                 SessionContext sessionContext = FrameworkUtils.getSessionContextFromCache(sessionId,
```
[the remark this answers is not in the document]
sadilchamishka: Good catch, fixed in fdf1350.

You are right that the short circuit was lost: the previous code resolved the session context first and only read from the database when it was non-null, so a session whose context is missing or expired cost no application info or metadata rows. Batching the read up front removed that.

The loop now resolves the contexts first into a LinkedHashMap, which also keeps the order of the caller's list, reads only the IDs that have a context, and then builds the result. The output is unchanged and the database work is a subset of what the previous revision of this PR did.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/dao/impl/UserSessionDAOImpl.java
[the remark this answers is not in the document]
sadilchamishka: The revision you checked, e00558d, predates the commit that added the logger, and the branch has since been squashed so that SHA is no longer on it. The current head is fdf1350, which has the logger declaration and the guarded DEBUG log reporting only the batch size before the two reads.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/dao/UserSessionDAO.java
```diff
@@ -94,6 +95,29 @@ default List<UserSession> getSessions(int tenantId, List<ExpressionNode> filter,
...
+        if (sessionIds == null) {
+            return sessions;
+        }
+        for (String sessionId : sessionIds) {
+            UserSession userSession = getSession(sessionId);
+            if (userSession != null) {
+                sessions.put(sessionId, userSession);
+            }
+        }
+        return sessions;
```
sadilchamishka: return empty list.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/store/SQLQueries.java
```diff
@@ -193,6 +193,15 @@ public class SQLQueries {
...
             "WHERE SESSION_ID = ?";
 
+    public static final String SQL_GET_APPS_FOR_SESSION_IDS = "SELECT SESSION_ID, SUBJECT, APP_ID FROM " +
+            "IDN_AUTH_SESSION_APP_INFO WHERE SESSION_ID IN (%s)";
+
+    public static final String SQL_GET_PROPERTIES_FROM_SESSION_META_DATA_FOR_SESSION_IDS = "SELECT SESSION_ID, " +
+            "PROPERTY_TYPE, VALUE FROM IDN_AUTH_SESSION_META_DATA WHERE SESSION_ID IN (%s)";
+
+    public static final String SQL_GET_PROPERTIES_FROM_SESSION_META_DATA_FOR_SESSION_IDS_H2 = "SELECT SESSION_ID, " +
+            "PROPERTY_TYPE, `VALUE` FROM IDN_AUTH_SESSION_META_DATA WHERE SESSION_ID IN (%s)";
```
sadilchamishka: use placeholder constant

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/dao/impl/UserSessionDAOImpl.java
```diff
@@ -385,14 +376,110 @@ private static int getDeleteChunkSize() {
...
+        return String.join(", ", Collections.nCopies(count, "?"));
+    }
+
+    /**
+     * Set the metadata properties of a session on the session object.
+     *
+     * @param userSession   Session to set the properties on.
+     * @param propertiesMap Properties of the session, which may be null when it has none.
+     */
+    private void setSessionProperties(UserSession userSession, Map<String, String> propertiesMap) {
```
sadilchamishka: refer the method as session metadata.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/dao/impl/UserSessionDAOImpl.java
```diff
@@ -78,59 +78,65 @@ public class UserSessionDAOImpl implements UserSessionDAO {
...
+                applicationsBySession.putAll(applicationsOfBatch);
             }
         } catch (DataAccessException e) {
             throw new SessionManagementServerException(
                     SessionMgtConstants.ErrorMessages.ERROR_CODE_UNABLE_TO_GET_SESSION,
                     SessionMgtConstants.ErrorMessages.ERROR_CODE_UNABLE_TO_GET_SESSION.getDescription(), e);
         }
-        return null;
+
+        applicationsBySession.values()
```
sadilchamishka: check whether we need.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/dao/impl/UserSessionDAOImpl.java
```diff
@@ -78,59 +78,65 @@ public class UserSessionDAOImpl implements UserSessionDAO {
...
+                .forEach(applications -> applications.removeIf(application -> application.getAppName() == null));
+
+        applicationsBySession.forEach((sessionId, applications) -> {
```
sadilchamishka: take above into this.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/store/SQLQueries.java
sadilchamishka: Fixed by - https://github.com/wso2/carbon-identity-framework/pull/8297/changes/47255abb5b546912cebaf24bf64aee184506a589

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/dao/impl/UserSessionDAOImpl.java
sadilchamishka: Fixed by - https://github.com/wso2/carbon-identity-framework/pull/8297/changes/47255abb5b546912cebaf24bf64aee184506a589

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/dao/UserSessionDAO.java
sadilchamishka: There are other implementations which haven't implemented it. Hence its better to have default implementation.

components/authentication-framework/org.wso2.carbon.identity.application.authentication.framework/src/main/java/org/wso2/carbon/identity/application/authentication/framework/dao/impl/UserSessionDAOImpl.java
sadilchamishka: Fixed.

